PR #8, opened 2020-09-17
Create zh_cn.json

qsefthuopq:


qsefthuopq:
What's the **main** here means?
I can't find it ingame.

Technici4n:
Thanks a lot for the translation!
The main is a mistake, but it never got removed from the JSON! We'll add a script to remove extra entries from the JSON!

PR #9, opened 2020-09-19
[bugfix] Fix assumption of PipeBlock that every BlockView is a World

romangraef:
This could cause crashes in conjunction with Torohealth which would oftentimes crash the client upon looking at a pipe.


Technici4n:
Nice, thanks!

PR #19, opened 2020-10-25
Russian language fix

Romz24:
I tested my previous version and found a lot of inaccuracies there. This fix should make the translation more accurate.


Azercoco:
I don't speak Russian so I'll trust you with that one

PR #37, opened 2020-12-15
colours of you, colour my life

Yoghurt4C:
just like the rainbow, burning bright
colours of you, colour my mind - you got it now
colours of you, colour me wild
and I just can't take off my eyes
colours of you, colour my heart just one more night
baby, make me feel alright

this pr:
1. changes the colour hexes of some materials
2. introduces a bit more diversity to the various material icons by adding `dull`, `gem`, `metallic`, `shiny` and `stone` icon sets - currently only targets items
3. changes the plate texture for metals to a less gradient-y version of the old "large plate", and scales them both down ~~so now the large plate isn't visually large, tell me if I should revert this 😹~~
grossly upscaled image as an example:
image


Yoghurt4C:
new changes:
1. added a builtin resourcepack with "alternate" textures based on gt6, which contains a lot of block overrides you may or may not want to make default
2. added icon sets for blocks
3. made automatic asset generation approximately 2 kroner slower by moving under/overlay checks to direct file queries
4. moved gt6-style plates to the builtin resourcepack, bringing back the ic2 plates as default
5. failed the spotless check

Technici4n:
Thanks a lot!

PR #46, opened 2021-01-27
Update KubeJS?

Neubulae:


Technici4n:
Thanks but it's not needed anymore now that I have removed KubeJS bundling.

Neubulae:
> Thanks but it's not needed anymore now that I have removed KubeJS bundling.

:wat: :hmpf:
who needs kjs bundling anyway:tiny_potato:

PR #63, opened 2021-05-23
Fix unspecified version in fabric.mod.json

deirn:
Setting the version after `processResources` makes it uses the old version value, which is `unspecified`, fixed it by moving the assignment to the top.
487xrSZ

I also made the commit hash as the version for master build actions. I just think it's a little nicer.
The version will defaults to `local` if built it locally without the env var.
2021-05-23 12_46_15-libs


Technici4n:
https://github.com/AztechMC/Modern-Industrialization/actions/runs/868059908

Seems to be working, thanks!

PR #75, opened 2021-07-13
DashLoader support

alphaqu:
This adds dashloader support through our api. Some visibility increasements and fixed cloth config being a non-existent version.


gradle.properties
```diff
@@ -10,7 +10,7 @@ org.gradle.jvmargs=-Xmx1G \
 # Fabric Properties
 	# check these on https://fabricmc.net/use
 	minecraft_version=1.17.1
-	yarn_mappings=1.17.1+build.1
+    yarn_mappings=1.17.1+build.1
```
Technici4n, line 13: Why

src/main/java/aztech/modern_industrialization/compat/dashloader/DashMesh.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public DashMesh(Mesh mesh) {
+        data = ((MeshImpl) mesh).data();
+    }
+
+    public Mesh toUndash() {
+        return MeshImplAccessor.create(data);
+    }
+}
```
Technici4n, line 48: Imo this should go into dashloader's API, instead of offloading internals access to other mods.

src/main/java/aztech/modern_industrialization/machines/init/MIMachineRecipeTypes.java
```diff
@@ -27,7 +27,6 @@
 import aztech.modern_industrialization.machines.recipe.CuttingMachineRecipeType;
```
Technici4n, line 27: This file was excluded from spotless, not how it got formatted it? This should be reverted.

src/main/java/aztech/modern_industrialization/machines/init/MultiblockMachines.java
```diff
@@ -23,6 +23,8 @@
  */
```
Technici4n, line 23: Should be reverted too.

src/main/java/aztech/modern_industrialization/machines/init/SingleBlockCraftingMachines.java
```diff
@@ -30,7 +30,6 @@
 import aztech.modern_industrialization.inventory.ConfigurableFluidStack;
```
Technici4n, line 30: Should be reverted too.

src/main/java/aztech/modern_industrialization/machines/models/MachineBakedModel.java
```diff
@@ -70,12 +71,18 @@ public boolean isVanillaAdapter() {
...
+            if (attachment instanceof MachineModelClientData clientData) {
+                MachineCasingModel casing;
+                if (clientData.casing == null) {
+                    casing = defaultCasing;
+                    System.out.println("defualt");
+                } else {
+                    casing = clientData.casing.mcm;
+                    System.out.println("client");
+                    System.out.println(clientData.casing.mcm.getMesh());
+                }
```
Technici4n, line 85: Debug prints. :smile:

Technici4n (changes requested):
Thanks a lot! Looks fine besides the unnecessary reformats, and the MeshImpl access (should go into dash imo).

src/main/java/aztech/modern_industrialization/machines/models/MachineBakedModel.java
alphaqu: oh god
i was about to take them out but then got a dm

gradle.properties
alphaqu: adverb
for what reason or purpose.
"why did he do it?"
adverb
(with reference to a reason) on account of which; for which.
"the reason why flu jabs need repeating every year is that the virus changes"
exclamation
1.
expressing surprise or indignation.
"why, that's absurd!"
2.
used to add emphasis to a response.
"‘You think so?’ ‘Why, yes.’"
noun
a reason or explanation.
"the whys and wherefores of these procedures need to be explained to students"

PR #77, opened 2021-07-15
Adjusted russian translation + created russian user guide for patchouli

flatron4eg:
There is a full russian translation for the mod, some lines are still in english (especially in achievements) due to wordplays which can't be translated well. Also I think you should use crowdin for translation purposes, it's a great tool.


Technici4n:
Thanks!

PR #103, opened 2021-08-18
Correctiting cuttingg machine descriptions

Orygeunik:
In english translation says that:
`"use lubricant" and no one word about water`
In source code also no means about water, only lubricant
https://github.com/AztechMC/Modern-Industrialization/blob/ed68f0a1072935b8e551f9d4b88e032f8bb41dc1/src/main/java/aztech/modern_industrialization/materials/recipe/StandardRecipes.java#L171

But in russian translation says (translated into english like as is):
> Rezhushchiy stanok dast vam luchsheye sootnosheniye dlya boltov, shesterenok, kolets i sterzhney. On kazhdyy raz ispol'zuyet nemnogo smazki. Poka vy mozhete ispol'zovat' tol'ko vodu, no v dal'neyshem lubrikanty uskoryat obrabotku.
213 / 5000
Результаты перевода
The cutting machine will give you the best ratio for bolts, gears, rings and rods. He uses a little lubricant every time. For now, you can only use water, but in the future, lubricants will speed up the processing. 

Pay attention to the third sentence


Azercoco:
I don't see any issue. I will merge. Thank you for pointing this.

PR #109, opened 2021-08-21
Modrinth auto publishing

m44ster:
I've tested that it works with a test mod on modrinth which is still a draft (I've also downloaded the jar just be sure)
grafik

even tho I haven't really touched the curseforge publishing part (I removed the Upload to Curseforge job and added it back tho),
I'm still worried that I broke something like I always am, so please verify that it still works

you will need to add a MODRINTH_API_KEY github secret


Technici4n:
Seems to work, thanks. Hopefully it will work when we publish a release. :smile:

PR #111, opened 2021-08-22
Add native support for macerator for silver ore

Orygeunik:
Hi!

Small change for runtime recipes which adding support to macerator recipes, like an original, for silver ore

Added separately for "tech reborn" and "industrial revolution".

Do you know, maybe yet another mods which have silver ore and which need add?

Before (not have available recipes for macerator):

image
image

After (available recipes for macerator):
image
image

---

And what are you think about extracting logic for runtime adding recipes per type/group/machine type in separate libs? For give an opportunity for other mod developers use this libs (and make some standard for registering, deduplicating and using runtime recipes)?


Azercoco:
I though we added it. Seems that we just forgot. Thank you for adding it.

Technici4n:
> And what are you think about extracting logic for runtime adding recipes per type/group/machine type in separate libs? For give an opportunity for other mod developers use this libs (and make some standard for registering, deduplicating and using runtime recipes)?

We don't really want to be maintaining this kind of library.
For unification please have a look at the AOF4 unification script that I wrote, and feel free to use and edit it to suit your needs.
For compat recipes, Fabric needs to provide a solution for it. This is currently in discussion... it will likely take a few weeks.

PR #114, opened 2021-08-23
Sync translating file and add missing recipes

Orygeunik:
Found lack of translate for barrels, also found some fluorite (which no in mod, where is it from?)

Before:
image

After:
image

Found lack of recipes for sphalerite ore:
image
And added it:
image

Also mod have existing the deep slate nikolite ore (which can received with steam drill, or pickaxe with silk touch), but it haven't any recipes for dust. I'll added first recipe for macerator like as for usally nikolite ore

image
image

Maybe is more correct add recipe into furnace for smelting deepslate nikolite ore into regular nikolite ore?


Technici4n:
Thanks!

PR #106, opened 2021-08-19
Small improve of ru translating

Orygeunik:
Fixed some problems like as:
Bord-free text:
image
image
image

Empty recipes image:
image

Broken markup:
image

Changed translate for word "machine".
At past "machine" translated like a "mechanism", in russian it's acceptably translate (it's may mean part of object), but i think more pretty and correct translates as "machine" (like as complex object, not part)

Also in some place rephrase russian text for more harmony and russian norm.
For example in russian in one paragraph not recommended use pronoun  twice in a row preferably alternate. It's a lexical repetition, and it's not very pretty.
As in English, it is not very good to start an answer with the phrase "because" and other

And so small attempt to translate achievements, but some references (easter eggs) very hard to translate, and in this case i saved original text (so my respect for some ref)

And so very-very-very-very (and more) hard...
Are you ready for this?

Changed translate for word "mixer".
At past "mixer" translated like... so it's hard... like as "mixer in bathroom", at now translated like "mixer for components"

>>> Place for brain explosion <<<

:-)

PR #117, opened 2021-08-24
Adding FAQ block into guide book

Orygeunik:
Hi, it's me again (tired already, yes?)

So, i looked on some commits, and so i figured it would be helpful:

Adding separated block for FAQ, with some helpers for users. At this moment QA was taken from discord channel (with same name, he-he)
Also removing duplicates of text from ru-text from pipes entity (this comment https://github.com/AztechMC/Modern-Industrialization/pull/106#issuecomment-902316067)

So, maybe you tell me what another QA exist which need to add?

_Perhaps titles are too jocky or not?_

Looks so:

General
image

EN
image
image
image

RU
image
image
image


Technici4n:
> I think so need fill some abstract "FAQ" and add link from main page to this.
> My point is that the page data can be used to expand the set of information, those. on the main pages there is a certain necessary minimum, and these pages expand and describe some details
>
> The concept is something like this:
Concentrated expression of thought in the main part, in the additional - a lot of description/examples and other things
>
> What do you think about it?

I'm not sure, I think it's better to group related information together as much as possible.

Orygeunik:
You are right, but i think so big info block instead of small was may scare away users.
Imagine the next case:
User starts read the guide book for find out smth about boiler, but he finds, a big "text wall", and his focus is dispersed  (he reading about everything, but not what he wanted, or he forggot what he want)

Azercoco:
I will close this PR. This discussion about Tooltips is moved on #303.

PR #122, opened 2021-09-06
Add 1.14ish coil texture

DcZipPL:
I made texture for Coil blocks that looks more 1.14ish.

This texture is under CC0 License

image


Azercoco:
Thank you for this texture. I will maybe slightly tweak the contrast.

PR #152, opened 2021-11-02
Add Arc Furnace aka Recycling machine

DcZipPL:
I add Arc Furnace machine that breaks down machines in to primitive components.

Arc Furnace requires graphite plates when crafting it. Graphite is blast-furnaced carbon dust. This material was made for consistency with other "Greg" like mods.

Arc Furnace currently don't have recipe autogen for any machine, at the moment you need to make recipe files manually.
I will add autogen if you want it.

GUI preview:
arc furnace gui


DcZipPL:
Currently don't merge it. I'm making fundamental changes to arc furnace. To make it more accessible for example removing graphite.

Azercoco:
I will close this issue because the Arc Furnace is way too late game to be useful for recycling. Discussion about early game recycling will be done in #304.

PR #178, opened 2021-12-12
Automating Furnaces and Blast Furnaces

Rakkis157:
Some recipes to fill in some holes in automation:

Furnace Assembler Recipe, for Bronze/Steel Boilers and Furnaces.
Blast Furnace Assembler Recipe, for SBF and EBF.
Blast Furnace Steel Plate crafting recipe.


src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/furnace.json
```diff
@@ -0,0 +1,17 @@
+{
+  "type": "modern_industrialization:assembler",
```
Azercoco, line 2: I would make this recipe in the packer instead

src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/furnace.json
```diff
@@ -0,0 +1,17 @@
...
+  "eu": 8,
+  "duration": 200,
+  "item_inputs": [
+    {
+      "tag": "minecraft:cobblestone",
```
Azercoco, line 7: the tag is #minecraft:stone_crafting_materials

Azercoco (changes requested):
Just apply the two little fixes and I will merge it. Thanks

Azercoco (approved):
Ok this is good ! Thank you ! Maybe we should add Steam Blast Furnace recipe in the packer ?

Rakkis157:
> Ok this is good ! Thank you ! Maybe we should add Steam Blast Furnace recipe in the packer ?

And the Coke Oven as well. I'll do that real quick.

Rakkis157:
It would be odd to have the blast furnace be an assembler recipe and the SBF be in the packer. Should I move the blast furnace to the packer as well, or have the SBF be in the assembler?

Azercoco:
> It would be odd to have the blast furnace be an assembler recipe and the SBF be in the packer. Should I move the blast furnace to the packer as well, or have the SBF be in the assembler?

No having the SBF in the packer is ok because it's in steam age where you don't have access to the assembler. It would not make sense to have the SBF in the assembler because when you unlock the latter, you have no use of the former anymore.

Rakkis157:
> > It would be odd to have the blast furnace be an assembler recipe and the SBF be in the packer. Should I move the blast furnace to the packer as well, or have the SBF be in the assembler?
> 
> No having the SBF in the packer is ok because it's in steam age where you don't have access to the assembler. It would not make sense to have the SBF in the assembler because when you unlock the latter, you have no use of the former anymore.

Also I was talking about moving the vanilla Blast Furnace to the packer. Should I do so?

Azercoco:
> > > It would be odd to have the blast furnace be an assembler recipe and the SBF be in the packer. Should I move the blast furnace to the packer as well, or have the SBF be in the assembler?
> > 
> > 
> > No having the SBF in the packer is ok because it's in steam age where you don't have access to the assembler. It would not make sense to have the SBF in the assembler because when you unlock the latter, you have no use of the former anymore.
> 
> Also I was talking about moving the vanilla Blast Furnace to the packer. Should I do so?

No I prefer that most electric age recipe stay in the assembler

Azercoco (approved):
Well done, and nice catch for the missing amethyst block in the packer !

PR #183, opened 2021-12-15
Create steam_quarry.json

Rakkis157:
Makes autocrafting the steam quarry and electric quarry possible.


Azercoco (approved):
I like the use of the Invar Rotary Blade

PR #184, opened 2021-12-15
Profession Blocks

Rakkis157:
Adding recipes for villager profession blocks, for players that want to build large trading stations or use those blocks for building. Recipes which use wood will have Creosote as a component, to give it a use outside of making lube.

### DIFF
--- a/src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/barrel.json
+++ b/src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/barrel.json
@@ -12,6 +12,12 @@
       "amount": 2
     }
   ],
+  "fluid_inputs": [
+        {
+            "fluid": "modern_industrialization:creosote",
+            "amount": 100
+        }
+  ],
   "item_outputs": [
     {
       "item": "minecraft:barrel",
--- a/src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/bell.json
+++ b/src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/bell.json
@@ -5,11 +5,7 @@
     "item_inputs": [
         {
             "item": "minecraft:stick",
-            "amount": 1
-        },
-        {
-            "item": "minecraft:stone_slab",
-            "amount": 2
+            "amount": 3
         },
         {
             "item": "modern_industrialization:bronze_plate",
--- a/src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/bell.json
+++ b/src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/bell.json
@@ -1,0 +1,35 @@
+{
+    "type": "modern_industrialization:assembler",
+    "eu": 8,
+    "duration": 200,
+    "item_inputs": [
+        {
+            "item": "minecraft:stick",
+            "amount": 1
+        },
+        {
+            "item": "minecraft:stone_slab",
+            "amount": 2
+        },
+        {
+            "item": "modern_industrialization:bronze_plate",
+            "amount": 1
+        },
+        {
+            "item": "modern_industrialization:bronze_curved_plate",
+            "amount": 2
+        }
+    ],
+    "fluid_inputs": [
+        {
+            "fluid": "modern_industrialization:creosote",
+            "amount": 100
+        }
+    ],
+    "item_outputs": [
+        {
+            "item": "minecraft:bell",
+            "amount": 2
+        }
+    ]
+}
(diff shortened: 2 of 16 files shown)


Rakkis157:
Kinda debating how to handle the cauldron recipe.
I feel like having it just use steel plates will be a source for frustration for newer players trying to autocraft steel gears.

Rakkis157:
Should be done now.

src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/barrel.json
```diff
@@ -12,10 +12,16 @@
       "amount": 2
     }
   ],
+  "fluid_inputs": [
+        {
+            "fluid": "modern_industrialization:creosote",
```
Azercoco, line 17: Not sure about using creosote as it's mean only for early game recipe ... maybe use lubricant instead

src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/brewing_stand.json
```diff
@@ -0,0 +1,21 @@
+{
+    "type": "modern_industrialization:assembler",
```
Azercoco, line 2: Maybe moving to packer ?

src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/composter.json
```diff
@@ -0,0 +1,23 @@
+{
+    "type": "modern_industrialization:assembler",
```
Azercoco, line 2: Moving this one to the packer to avoid conflicting with the barrel

Azercoco (changes requested):
There is a lot of potential recipe conflict... I know slot locking exists but the less it is needed, the better. I would move nearly every of these recipe to the packer instead where the possibility of conflict is more limited by the number of slots

src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/barrel.json
Rakkis157: Creosote was chosen because it is used irl to treat wood so it is more resilient to termites etc.  That, and I figured that since the people who are most likely to use these recipes are builders, they might still be using Coke Ovens for Coke Blocks and/or Dye production, and have excess Creosote lying around.

src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/composter.json
Rakkis157: I'm adding seeds to the composter recipe, because only moving to the packer creates a conflict with the pressure plate recipe.

Rakkis157:
The recipes for Smithing Table, Loom, Fletching Table and Cartography Tables already don't clash with anything, as far as I can tell. Same goes for the Smoker.

src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/brewing_stand.json
Rakkis157: Done.

Rakkis157:
Added a bronze ring and bronze bolt to the bell recipe, to reduce chance of recipe conflict.

Rakkis157:
Composters, Bookshelves and Lecterns shouldn't have any issues now, as they all require a unique component. (Seeds, Books, Bookshelves)

Rakkis157:
Also changed the Cauldron recipe, to prevent conflicts with the bucket.

src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/assembler/bell.json
```diff
@@ -0,0 +1,35 @@
...
+        {
+            "item": "modern_industrialization:bronze_ring",
+            "amount": 1
+        },
+        {
+            "item": "modern_industrialization:bronze_plate",
+            "amount": 1
+        },
+        {
+            "item": "modern_industrialization:bronze_bolt",
```
Technici4n, line 15: Bolts are not meant to be used in the assembler (they are replaced by soldering alloy).

src/main/resources/data/modern_industrialization/recipes/vanilla_recipes/packer/cauldron.json
```diff
@@ -0,0 +1,31 @@
...
+  "type": "modern_industrialization:assembler",
+  "eu": 8,
+  "duration": 200,
+  "item_inputs": [
+    {
+      "item": "modern_industrialization:steel_curved_plate",
+      "amount": 4
+    },
+    {
+      "item": "modern_industrialization:steel_bolt",
```
Technici4n, line 11: Same comment about bolts.

Technici4n (changes requested):
Looks really good besides the bolt issue! Thanks.

Azercoco:
Closed for the same reasons as #188

PR #185, opened 2021-12-15
Aquatic Recipes

Rakkis157:
A bunch of recipes around prismarine and coral.

Prismarine Crystal recipe is based on the process for making Aquamarine irl, but using salt water (kinda) instead of pure water.

Coral is being "grown" in the Chemical Reactor.
~~Wish there was an aquatic farm machine~~


Azercoco (approved):
Ok this is good ! Ty

PR #198, opened 2021-12-22
Fix spelling of disintegrate

NiChrosia:
Fix the spelling of 'desintegrte' in the quantum sword advancement description.


Technici4n:
Thanks. I think the added trailing line return is making the formatter fail? Can be fixed with `gradlew spotlessApply`.

PR #205, opened 2021-12-26
temp fix for outdated Chinese Guidebook

jecowa:
Updates some item names so the Chinese guidebook shows pages. (Still needs updating, though)


Technici4n:
Thanks!

PR #210, opened 2021-12-28
Google translation of Guidebook to Chinese

jecowa:
here's a google translation if you want


Technici4n:
I don't really see the point of a google translation... seems like it would be quite low quality?

jecowa:
A Google translation might be an easier starting point. Someone who knows Chinese can copy-edit the Google translation without knowing English or running it all through Google translate themself.

Currently the Chinese Guidebook is blank. Until someone can do a good translation, the options to make it useable are:
 * delete the Guidebook's Chinese translation so it falls back the English version,
 * fix the current translation so that it is useable (though may have some outdated information),
 * or use machine translation.

KaneHart:
If it's released with a Broken Book then the question is why was that ever a thing before we jump on the guy who google translated it to be fair. Weather or not that was a mistake / bug. We should fix it one way or another including just putting it to the English one till someone can properly translate it.

Technici4n (changes requested):
I'm fine with google-translated pages, but that should only be done for pages which didn't have a manual translation before.

jecowa:
I replaced the Google-translated lines with the human-translated lines where available. Made one small change to the human-translation to match the current and say "20 times cheaper" instead of "40 times cheaper". Also edited the Google-translated lines to better match the human-translated lines and say "sneak+right click" instead of "shift + right click". And removed some extra spaces.

Technici4n:
That looks nice, thank you! Will be in the next release.

PR #223, opened 2022-01-10
Update OrePart.java to allow ore gen in non-vanilla biomes and datapack biomes

Miku0Sama:
I added a custom biome selection context that, when used in the "add feature" function in place of fabric biome selectors, allows all ores to be generated in any overworld biome regardless of how the biome is added to minecraft.

### DIFF
--- a/src/main/java/aztech/modern_industrialization/materials/part/OrePart.java
+++ b/src/main/java/aztech/modern_industrialization/materials/part/OrePart.java
@@ -40,8 +40,10 @@ import java.io.IOException;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.function.Predicate;
 import net.devtech.arrp.json.loot.*;
 import net.fabricmc.fabric.api.biome.v1.BiomeModifications;
+import net.fabricmc.fabric.api.biome.v1.BiomeSelectionContext;
 import net.fabricmc.fabric.api.biome.v1.BiomeSelectors;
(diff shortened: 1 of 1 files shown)


Technici4n:
If this is the fix then it looks like Terralith fails to register its biomes in the OVERWORLD category and it's up to them to fix it.

Miku0Sama:
It's actually an issue with Datapacks not Terralith. For whatever reason, biomes added via data don't get registered to any dimension categories. Many mods add biomes via datapacks now and so this is the solution that AE2, Bewtichment, Gobber, and many others use to circumvent this issue as there is nothing that the biome creators can do about Mojangs implementation of Datapacks. This solution also allows the ores to generate in other non-vanilla dimensions such as mining worlds. You can also implement biome and dimension white/black listing via configs with this setup. Though I understand if that isn't something you want in this mod.

Bit of a digression but I think this might be an indication of an issue with the default Fabric selectors as Datapack biomes do have a category but are often given a sub-category such as "forest" or "mesa" which are part of the overworld category but aren't the overworld category directly. Though I'm not entirely sure as I'm not that experienced with Fabric yet so maybe you might know more about that.

Miku0Sama:
> there is nothing that the biome creators can do about Mojangs implementation of Datapacks.

At least that's what the datapack creators have said...

Technici4n:
Just confirmed that this should be a good fix by looking at what AE2 does. Indeed, it would possibly be worth PR'ing a fix to fabric.

Technici4n:
Fixed in latest Fabric API.

PR #224, opened 2022-01-11
Added Wax

Rakkis157:
Wax item, plus a couple basic recipes.
Have a few other stuff that I'll do on a later date.
Not sure if should make the wax item work like the honeycomb. Having it not work like that would give more incentive to automate.

image
image


Azercoco (changes requested):
Thanks for the PR but shouldn't the wax implement the Farbric waxing API ?

Azercoco:
Implemented within db289a33023afaefb849613667f26b63b23fed62

PR #264, opened 2022-02-24
Add filling and unfilling machines

maksalees:
This PR adds a Filling Machine and an Unfilling Machine which can be used to automatically pour and empty fluids into buckets. Also added achievements for crafting them.

# Screenshots

# Crafts
## Filling Machine
Screenshot_20220224_161816
## Unfilling Machine
Screenshot_20220224_161825


Azercoco:
The idea is nice but I think it would be better if the filling/unfilling machines were using the full Fluid AP so they could for example be used to fill/unfill tank or container from other mod

maksalees:
> The idea is nice but I think it would be better if the filling/unfilling machines were using the full Fluid AP so they could for example be used to fill/unfill tank or container from other mod

I just looked at the Fluid Transfer API code and didn't see any way to get all registered fluid storages. Is it possible?

Technici4n:
No, and that's not how it should work. There is no reason to use a recipe system for this, it's much better to just try filling whatever item is in the input slot.

k3nw4y:
does it has any functionality? if its just for autocrafting, its better to use AE2s new fluid crafting, iirc MI has no use for liquid in buckets/containers

Azercoco:
I will close this issue. Discussion moved to #305.

PR #288, opened 2022-05-06
Added ingot->plate for TR compressor for missing ingots

Ayutac:
Untested.


Technici4n:
Superseded by #306 that refactors the material system a bit to make datagen nicer. Thanks for the initial implementation!

PR #293, opened 2022-05-10
fix allowNormalOperation not being corrected on server load

aria1th:
should not produce any side effects


Technici4n:
thanks for the PR, but I would prefer properly fixing the multiblock shape detection system cache instead of working around the issue

PR #294, opened 2022-05-14
Improve Simplified Chinese (zh_cn) Translation

MineCommanderCN:
The original Simplified Chinese translation was terrible, with many inconsistent nouns, outdated keys and machine translation traces. So I changed some texts in zh_cn.json, removed all deprecated keys and unified all specialized nouns. Also re-translated guide book completely.


Azercoco:
I will merge it for now but we are reworking the translations so a lot of changes may be needed in the near future.

PR #380, opened 2022-09-25
Accept all tagged barrels when crafting MI barrels

### LINKED ISSUE #371: Barrel variants not working in the recipe of metal barrels.
Barrel variants such as the ones from Better Nether do not work in the recipes for the metal barrels (steel barrel, aluminium barrel etc.).
This can be fixed by using '#c:barrels/wooden' tag in the recipes instead of 'minecraft:barrel' item reference.

JustinHuPrime:
Fixes #371


Technici4n:
Thanks, but we already have the fix in a WIP branch that is includes a larger barrel rework: https://github.com/AztechMC/Modern-Industrialization/commit/7c17a439308a4cfb373f969e7c51a2677bd20e4a.

Note that we're using `#c:wooden_barrels` instead of `#c:barrels`.

PR #405, opened 2022-11-27
Fix #404: add tolerance for null return from getting held item

### LINKED ISSUE #404: Crash while rendering electric quarry
I encountered a null pointer exception, apparently in code trying to render an electric quarry - it was querying if I was holding a wrench; I can't recall what the exact circumstances around it were, but I wasn't interacting with the machine, and I might have scrolled past the wrench, but I'm not certain.

I've attached the relevant log file - [2022-11-26-5.log](https://github.com/AztechMC/Modern-Industrialization/files/10097773/2022-11-26-5.log) - and the relevant crash report - [crash-2022-11-26_23.44.54-client.txt](https://github.com/AztechMC/Modern-Industrialization/files/10097776/crash-2022-11-26_23.44.54-client.txt)

The log file points at [these lines](https://github.com/AztechMC/Modern-Industrialization/blob/7713aa252124d1ae3145d3705449d78afeaf7fc7/src/client/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBER.java#L89-L92), and after remapping the crash report's names, it looks like `player.getMainHandItem()` may return null.

JustinHuPrime:
Closes #404


JustinHuPrime:
Amended commit to fix SpotlessJava build failure

Technici4n:
No, this makes no sense. Stacks are not supposed to have `null` items.

JustinHuPrime:
I agree that stacks shouldn't have `null` items, but I can't think of any other reason for the crash report in #404.

Specifically, here's the head of the stacktrace, with the obfuscated identifiers remapped:
```
java.lang.NullPointerException: Cannot invoke "net.minecraft.world.item.Item.builtInRegistryHolder()" because the return value of "net.minecraft.world.item.ItemStack.getItem()" is null
	at net.minecraft.world.item.ItemStack.is(ItemStack.java:236)
	at aztech.modern_industrialization.machines.multiblocks.MultiblockMachineBER.isHoldingWrench(MultiblockMachineBER.java:91)
```

Ah, oops - taking a second look at that stacktrace, my fix wouldn't actually fix anything.

Technici4n:
I remember seeing a message by some Iris dev saying that this issue can happen when the `ItemStack` is being modified in another thread.

PR #413, opened 2022-12-08
Fix book's diesel generator throughput stats

JustinHuPrime:
The stats for diesel generators given in the Patchouli book are out-of-sync with the actual stats:

- The regular diesel generator produces 256 EU/t, not 128
- The turbo diesel generator produces 1024 EU/t, not 512


Technici4n:
Good spot, thanks!

PR #454, opened 2023-03-24
Fix incorrect tags on dye recipes generation

londocr:
2 tags were wrong:

- the correct tag for wool carpets is `#minecraft:wool_carpets`
- the correct tag for shulker boxes is `#c:shulker_boxes`(in plural)


piousp (approved):
Thank you

PR #456, opened 2023-03-26
update Modrinth Auto publishing

m44ster:
I know there are a lot of commits, most of them are just to test things and got reverted in another commit, I hope thats fine.


m44ster:
btw that (see below) means its gonna skip the modrinth publishing if there is no token and you can only see that in the log which might be bad, I can change that if you want.
```
tasks.modrinth.onlyIf {
    System.getenv("MODRINTH_API_KEY") && !version.endsWith("-SNAPSHOT")
}
```

Technici4n:
Thanks!

PR #464, opened 2023-04-09
Adding custom steam machine overclock capabilities

PurelySimple:
### What

- Ability to create new steam machine overclocking catalysts or disable the default gunpowder one
- Ability to create steam multiblock machines in KubeJS
- Ability to further configure REI with options to set workstations, steam, or recipe filtering
- Updated documentation explaining how to use the new features

### Steam multiblock
Steam multiblocks can now be created in KubeJS with the new `simpleSteamCraftingMultiBlock` method. Its exactly identical to `simpleElectricCraftingMultiBlock` method so its easy to switch between and doesn't require separate documentation to explain.
Example:
```
    event.simpleSteamCraftingMultiBlock(
        "Basic Steam", "basic_steam", CUSTOM_MULTIBLOCK, testShape,
        // REI parameters
        event.progressBar(77, 33, "arrow"),
        itemInputs => itemInputs.addSlots(56, 35, 1, 2), itemOutputs => itemOutputs.addSlot(102, 35),
        fluidInputs => fluidInputs.addSlot(36, 35), fluidOutputs => fluidOutputs.addSlot(122, 35),
        // Model
        "heatproof_machine_casing", "pyrolyse_overlays", true, false, false
    );
```
image

### Optional Config parameter
To differentiate required parameters from optional ones, a new config lambda parameter was added at the end of the following RegisterMachineEventJS methods: `craftingSingleBlock`, `simpleElectricCraftingMultiBlock`, and `simpleSteamCraftingMultiBlock`. To make them backwards compatible with existing modders scripts, existing methods pass a default `config -> {}` to the new methods signaling no additional configuration is needed.

The `craftingSingleBlock` method is setup to handle both steam and electric machines based on tiers selected so a new `SingleBlockCraftingMachines.Config` class was created to house future configuration. For this PR, steam overclocking is the first feature to use it.

The `simpleElectricCraftingMultiBlock` and `simpleSteamCraftingMultiBlock` methods also follow the same pattern of having a non config lambda version and one with. While not necessary for backwards compatibility with the new steam method, symmetry was kept for consistency. Both crafting multiblock methods use the same config object `ExtraMachineConfig.CraftingMultiBlock` so switching only the type (ie Electric to Steam) still works. This means on an electric crafting multiblock it's possible to call config settings that work only on a steam machine (ie steamCustomOverclock) but doing so doesn't actually affect the electric machine. The upside is that shared code (ie reiExtra) can work for either machine type without duplicating efforts.

### Custom Overclock
Steam machines, both single block and multiblock variants, can now be configured to have custom overclock items instead of just gunpowder. It's also possible to disable overclocking with the feature as well.

Example:
```
    event.simpleSteamCraftingMultiBlock(
        "Advanced Steam", "advanced_steam", CUSTOM_MULTIBLOCK, testShape,
        // REI parameters
        event.progressBar(77, 33, "arrow"),
        itemInputs => itemInputs.addSlots(56, 35, 1, 2), itemOutputs => itemOutputs.addSlot(102, 35),
        fluidInputs => fluidInputs.addSlot(36, 35), fluidOutputs => fluidOutputs.addSlot(122, 35),
        // Model
        "heatproof_machine_casing", "pyrolyse_overlays", true, false, false,
        // Extra config
        config => config
            .steamCustomOverclock(1.5, {'minecraft:glowstone_dust': 1200})
    );
```
This example creates a steam multiblock that can be right clicked with glowstone dust to provide a 1.5x speed increase for 1200 ticks. A custom tooltip is generated on the machine to let the player know it's possible.
image

The implementation works by creating a new shared `IOverclockComponent` interface that is shared by the existing `GunpowderOverclockComponent` class and the new `CustomOverclockComponent`class. Existing steam machines default to using the GunpowderOverclockComponent while the CustomOverclockComponent is only used by the new KubeJS functions. The formula for deciding what the recipe overclock level was hoisted from the machine layer into the interface so the implementation has more control and less parameters had to be exposed out of it.

### REI Extra Config Options
Custom multiblock machines can now access more REI options. By default all steam crafting multiblock machines show steam as the default energy type. Its now possible to add multiple workstations, change energy types, and filter recipes based on a condition. All of this is done through a new `reiExtra` method on the config object.

Example:
```
        config => config.reiExtra(rei => rei.workstations('advanced_steam', 'steam_quarry').steam(true))
```

### New Documentation
The documentation was updated in a few parts to make it known that steam crafting multiblocks can be created, steam machines overclock functionality can be changed, and REI can be further configured.

image


docs/ADDING_MACHINES.md
```diff
@@ -78,14 +78,26 @@ MIMachineEvents.registerMachines(event => {
       items => items.addSlots(42, 27, 3, 3).addSlots(139, 27, 1, 3), fluids => {},
       /* MODEL CONFIGURATION */
       // front overlay?, top overlay?, side overlay?
-      true, true, false,
```
Technici4n, line 81: It's good practice to leave an extra comma (in the languages that allow it). :wink:

docs/ADDING_MACHINES.md
```diff
@@ -78,14 +78,26 @@ MIMachineEvents.registerMachines(event => {
...
 ```
+If bronze or steel is in the list of tiers, the machine behavior defaults to allow gunpowder to double the speed of the machine. You can change this behavior by adding a config lambda to the end.
 
-## Adding an electric multiblock crafting machine
-These are the standard electric multiblock machines, such as the pressurizer, the electric quarry, etc...
+Example:
+```
+// steamCustomOverclock(Speed multiplier, Map of item or block to duration in ticks) 
+config => config.steamCustomOverclock(3, {"minecraft:redstone": 400, "minecraft:redstone_block": 3600 })
+```
```
Technici4n, line 91: What if we want to allow different speed multipliers depending on the material? I feel like that would be a useful addition.

docs/ADDING_MACHINES.md
```diff
@@ -146,6 +167,41 @@ With KubeJS, the texture can go in the following folders respectively:
...
+    );
+})
+```
+Methods that config exposes:
+* `steamCustomOverclock(1.5, {"minecraft:redstone": 400, "minecraft:redstone_block": 3600})`
+  * Only on steam multiblocks, allows changing the overclock amount and duration for the given items or blocks.
+  * This example adds a 1.5x speed multiplier for 400 ticks when redstone is applied or 3600 ticks for a redstone block.
+  * Use `steamCustomOverclock(1, {})` to disable the default gunpowder overclocking behavior.
+* `reiExtra(rei => {})`
+  * Allows extra / advanced REI configuration. Methods can be chained.
```
Technici4n, line 197: Would be nice to mention the default values. :)

src/main/java/aztech/modern_industrialization/compat/kubejs/machine/ExtraMachineConfig.java
```diff
@@ -0,0 +1,76 @@
...
+            var catalysts = new ArrayList<CustomOverclockComponent.Catalyst>();
+
+            for (var entry : object.entrySet()) {
+                catalysts.add(new CustomOverclockComponent.Catalyst(new ResourceLocation(entry.getKey()), entry.getValue().getAsInt()));
+            }
+            steamOverclockComponent = new CustomOverclockComponent(multiplier, catalysts);
+            return this;
+        }
+
+        public CraftingMultiBlock reiExtra(Consumer<MultiblockMachines.Rei> consumer) {
```
Technici4n, line 71: Maybe it would be wiser to store them in a list, and apply them all, in case people call `reiExtra` multiple times.

src/main/java/aztech/modern_industrialization/compat/kubejs/machine/RegisterMachinesEventJS.java
```diff
@@ -97,16 +125,98 @@ public void simpleElectricCraftingMultiBlock(
...
+        extraConfig.accept(config);
+
+        simpleMultiBlock(englishName, internalName, recipeType, multiblockShape, progressBar,
+                itemInputPositions, itemOutputPositions, fluidInputPositions, fluidOutputPositions,
+                controllerCasingName, overlayFolder, frontOverlay, topOverlay, sideOverlay,
+                bep -> new SteamCraftingMultiblockBlockEntity(bep, internalName, multiblockShape, recipeType, config.steamOverclockComponent),
+                config.reiConfig);
+    }
+
+    public void simpleMultiBlock(
```
Technici4n, line 197: Should be `private`

src/main/java/aztech/modern_industrialization/machines/components/CustomOverclockComponent.java
```diff
@@ -0,0 +1,145 @@
...
+import net.minecraft.core.particles.ParticleTypes;
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.network.chat.Component;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.InteractionHand;
+import net.minecraft.world.InteractionResult;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+
+public class CustomOverclockComponent implements IOverclockComponent {
```
Technici4n, line 41: This should be merged with the gunpowder component as it's a generalization of it.

src/main/java/aztech/modern_industrialization/machines/init/SingleBlockCraftingMachines.java
```diff
@@ -179,11 +193,14 @@ public static void registerMachineTiers(String englishName, String machine, Mach
             MachineGuiParameters builtGuiParams = guiParamsBuilder.build();
             String id = prefix + "_" + machine;
 
+            var overclockComponent = (extraConfig.steamOverclockComponent != null) ? extraConfig.steamOverclockComponent
```
Technici4n, line 196: This feels wrong - the config object should be properly constructed before it's passed to this function. (See below)

Technici4n (changes requested):
Very good start, a few things to change and it will be good for me.

docs/ADDING_MACHINES.md
PurelySimple: That would be interesting but would change the default behavior a lot more. The default behavior of gunpowder is to allow stacking the tick duration as you add more gunpowder. If each catalyst were to affect the speed directly then it would have to overwrite the existing. Take for example the following hypothetical format
```
config.steamCustomOverclock({
	"minecraft:redstone": {
		"multiplier": 2,
		"ticks": 400
	},
	"minecraft:glowstone_dust": {
		"multiplier": 3,
		"ticks": 1000
	}
})
```
The intent is that glowstone is the more powerful catalyst and lasts longer than redstone.

If I right click with redstone in my hand twice, resulting in 800 ticks at 2x multiplier, what should happen if I right click with glowstone dust? Should it be at 3x multiplier with 1800 ticks allowing me to accumulate ticks from a different catalyst on a more powerful one? Or should it override and become 3x multiplier with 1000 ticks making me lose all my catalyst usages on the redstone?

What if I were to reverse the process and add a glowstone dust first and then redstone. Do I lose the 3x multiplier and reset back to 2x with 400 ticks? That would feel bad.

It's possible to implement either way but it would require retraining the player from the default behavior that throwing more catalysts in gets you more time.

PurelySimple: Perhaps another implementation could be to prevent the user from inserting a new catalyst until the existing one finishes? All of these could be correct depending on the design.

src/main/java/aztech/modern_industrialization/compat/kubejs/machine/RegisterMachinesEventJS.java
PurelySimple: Sure. I was thinking maybe it would make some ones life easier who was knowledgeable in kubejs reflection and would allow them to create custom EBF, boilers, etc more easily. They would have to be looking at the source to find it anyway but if the plan is to add things for that in a more friendly way (ie documented) then it probably should be private.

src/main/java/aztech/modern_industrialization/machines/components/CustomOverclockComponent.java
PurelySimple: It's certainly possible to make gunpowder just configuration in the CustomOverclockComponent since it contains more functionality. I only left it separate so there were less chance of breaking existing MI gunpowder behavior. From a maintainability standpoint it would be better to have one. If I remove the GunpowderOverClockComponent, should I remove the IOverclockComponent interface and rename CustomOverClockComponent to just OverclockComponent?

src/main/java/aztech/modern_industrialization/compat/kubejs/machine/ExtraMachineConfig.java
PurelySimple: Right now if you called reiExtra multiple times only the last lambda would be executed. The point of the consumer passing an rei object is that you can chain commands on the object to avoid extra typing (ie builder pattern). If the consumers are stored in a list and applied in order then the last setting still wins.

Is it standard behavior elsewhere in the API to allow calling a builder consumer multiple times? IMO, I'd rather error if they use the method twice and tell them to chain commands and use the builder pattern properly but it's up to you.

docs/ADDING_MACHINES.md
Technici4n: Ah, I see what you mean, that's a very good point. I'll think more about it.

src/main/java/aztech/modern_industrialization/compat/kubejs/machine/ExtraMachineConfig.java
Technici4n: I agree that one should:tm: use the builder pattern properly, but from the documentation it would be easy to copy/paste multiple `.reiExtra` calls. Sadly not everyone is familiar with programming, and we should try to accommodate for that when we can.

For example, one might copy both `reiExtra(rei => rei.workstations('pyrolyse_oven', 'steam_quarry'))` and `reiExtra(rei => rei.steam(false))`.

src/main/java/aztech/modern_industrialization/machines/components/CustomOverclockComponent.java
PurelySimple: Minor breaking change but GunpowderOverclockComponent writes "overclockGunpowderTick" NBT where as the new OverlockComponent writes "overclockTicks" NBT which will result in machines losing their gunpowder overclock ticks when upgrading to the new version. Is this an issue?

Technici4n: Yes, the format should be kept compatible. In the reading method, you can check for both. :Wink:

Make sure to add a `// TODO 1.20: Remove upgrade code`.

src/main/java/aztech/modern_industrialization/compat/kubejs/machine/ExtraMachineConfig.java
```diff
@@ -43,33 +41,32 @@ public CraftingSingleBlock(SingleBlockCraftingMachines.Config config) {
...
             return this;
         }
     }
 
     public static class CraftingMultiBlock {
-        public IOverclockComponent steamOverclockComponent = new GunpowderOverclockComponent();
-        public Consumer<MultiblockMachines.Rei> reiConfig = rei -> {
-        };
+        public OverclockComponent steamOverclockComponent = OverclockComponent.createDefaultGunpowderOverclock();
+        public ArrayList<Consumer<MultiblockMachines.Rei>> reiConfigs = new ArrayList<Consumer<MultiblockMachines.Rei>>();
```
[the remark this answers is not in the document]
Technici4n: (Following Java conventions the field should be a `List` even if the implementation is an `ArrayList`).

docs/ADDING_MACHINES.md
Technici4n: I think we should keep track of all the remaining ticks for each possible multiplier, and we apply the fastest multiplier first.

For example if we have 10 ticks of x2 and we suddenly add 5 ticks of x3, it will be x3 for the first 5 ticks and then x2 for the remaining 10 ticks.

This can be done using a `SortedMap` (`TreeMap` implementation), as it allows getting the last entry of the map easily.

PurelySimple: It looks like it will have to be a `NavigableMap` since `SortedMap.lastKey()` throws if the map is empty (not something we should catch in tick) but `NavigableMap.lastEntry()` will return null if empty. It's assumed that multiplier entries with no ticks have to be removed from the map so getting the last entry updates to the next highest multiplier. Bear in mind lastKey or lastEntry with a TreeMap is still O(log N) called in tick.

Since a lookup for the current multiplier and a deduction in its tick count happens every tick and additions to ticks on a specific multiplier only happen when a catalyst is inserted (very infrequent) it would be more performant to use an `ArrayList`. Walk it on update or insert for O(N) to prevent duplicates and keep it sorted. Getting the highest multiplier in tick by just grabbing and updating or removing the last element is O(1).

Either implementation works but its your call. There's probably far worse things than O(log N) at 20hz going on in Minecraft to really matter. 🥲

PurelySimple:
Updated documentation and changed the data structure to allow specifying different multipliers.
image

docs/ADDING_MACHINES.md
Technici4n: Ah yes array list works fine too of course. Whatever is fine by me :)

> Also with this addition it likely means the multiblock machine needs to display its current multiplier and remaining ticks to even know if its working correctly. Single block machines show the remaining time but not the multiplier.

We don't need to show too much info, this is an edge case after all. The current amount of info is probably fine.

docs/ADDING_MACHINES.md
```diff
@@ -146,6 +178,65 @@ With KubeJS, the texture can go in the following folders respectively:
...
+```
+Methods that config exposes:
+
+### `config.steamCustomOverclock()`
+* Only on steam multiblocks, allows changing the overclock amount and duration for the given items or blocks.
+* Use `steamCustomOverclock({})` to disable the default gunpowder overclocking behavior.
+* If multiple items are contributing to the machines overclock, the highest multiplier is processed first.
+* Items with the same multiplier value have their ticks summed together
+
+Example:
```
Technici4n, line 210: Can we put this explanation above in the normal machine section, and only refer to it here. Something like

```
* Only on steam multiblocks, allows changing the overclock amount and duration for the given items or blocks.
* See above, same as single block machines.
```

This is to avoid duplication of info.

src/main/java/aztech/modern_industrialization/compat/kubejs/machine/ExtraMachineConfig.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        public CraftingSingleBlock steamCustomOverclock(JsonObject object) {
+            var catalysts = new ArrayList<OverclockComponent.Catalyst>();
+
+            for (var entry : object.entrySet()) {
+                var catalystName = entry.getKey();
+                var parameters = entry.getValue().getAsJsonObject();
+                if (!parameters.has("multiplier")) {
+                    throw new RuntimeException(String.format("Expected multiplier key in object for catalyst %s", catalystName));
```
Technici4n, line 51: It will be cleaner to use the `GsonHelper` methods, for example `GsonHelper.getAsDouble(object, "multiplier")` which will handle the error and the conversion.

src/main/java/aztech/modern_industrialization/machines/components/OverclockComponent.java
```diff
@@ -0,0 +1,189 @@
...
+import net.minecraft.world.InteractionHand;
+import net.minecraft.world.InteractionResult;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+
+public class OverclockComponent implements IComponent {
+
+    private List<Catalyst> catalysts;
+
+    private NavigableMap<Double, MutableTickCount> TickMap = new TreeMap<>();
```
Technici4n, line 45: Naming should be `tickMap`, otherwise usages look like static method calls.

src/main/java/aztech/modern_industrialization/machines/components/OverclockComponent.java
```diff
@@ -0,0 +1,189 @@
...
+import net.minecraft.network.chat.Component;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.InteractionHand;
+import net.minecraft.world.InteractionResult;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+
+public class OverclockComponent implements IComponent {
+
+    private List<Catalyst> catalysts;
```
Technici4n, line 43:
```suggestion
    private final List<Catalyst> catalysts;
```

src/main/java/aztech/modern_industrialization/machines/components/OverclockComponent.java
Technici4n, line 45: Whether you want to use a map or a list is up to you.

src/main/java/aztech/modern_industrialization/machines/components/OverclockComponent.java
```diff
@@ -0,0 +1,189 @@
...
+    public List<Component> getTooltips() {
+        var tooltips = new ArrayList<Component>();
+        for (Catalyst catalyst : catalysts) {
+            Component catalystName;
+            if (Registry.BLOCK.containsKey(catalyst.resourceLocation)) {
+                catalystName = Component
+                        .translatable("block.%s.%s".formatted(catalyst.resourceLocation.getNamespace(), catalyst.resourceLocation.getPath()));
+            } else if (Registry.ITEM.containsKey(catalyst.resourceLocation)) {
+                catalystName = Component
+                        .translatable("item.%s.%s".formatted(catalyst.resourceLocation.getNamespace(), catalyst.resourceLocation.getPath()));
```
Technici4n, line 143: You should get the `item` and call `item.getName` which will handle these cases correctly (and will also take into account mods overriding it).

Technici4n (changes requested):
Still a few things to change, but getting very close. :)

src/main/java/aztech/modern_industrialization/machines/components/OverclockComponent.java
PurelySimple: It looks slightly different with the brackets but Im guessing that's intended?
image

Technici4n: I believe you need `getHoverText`?

PurelySimple: image

Issue #506, opened 2023-06-02
Config option to disable default trades


TheStaticVoid:
[Discord conversation](https://discord.com/channels/754461970870173767/760125612126896196/1114252463613169805)

PR #510, opened 2023-06-05
Add config option to disable Industrialist villager

TheStaticVoid:
This should be everything needed to add the config option for #506 . The changes use the config setting to block the registry of the villager profession, villager trades, and the forge hammer as a POI.

I'm still new to MC modding, so please let me know if I overlooked or messed up on anything!


src/main/java/aztech/modern_industrialization/MIVillager.java
```diff
@@ -41,13 +41,16 @@
...
+    public static final PoiType POI_TYPE = (MIConfig.getConfig().disableMIVillagers) ? null : PointOfInterestHelper.register(ID, 1, 1, MIBlock.FORGE_HAMMER.asBlock());
     public static final ResourceKey<PoiType> POI_KEY = ResourceKey.create(Registry.POINT_OF_INTEREST_TYPE_REGISTRY, ID);
 
     public static final VillagerProfession PROFESSION = new VillagerProfession(ID.toString(), e -> e.is(POI_KEY),
             e -> e.is(POI_KEY), ImmutableSet.of(), ImmutableSet.of(), SoundEvents.VILLAGER_WORK_TOOLSMITH);
 
     public static void init() {
+        if (MIConfig.getConfig().disableMIVillagers) return;
+
+
```
James103, line 53: Given the rest of the code here, one of the two empty lines here aren't needed.

Technici4n (changes requested):
Do you really want to disable the villager itself? I would prefer disabling the trades only. Does the tag loading not print an error now? (POI types need to be tagged to work for villagers).

TheStaticVoid:
I can change it so the villager profession and POI registration happens regardless, and then escape the init method before the trades are added if that's what you prefer. As for the tag error I fixed that in the second commit, unless you're referring to something different?

Technici4n:
Yes I think that would be better. Datagen should not depend on a config option, otherwise everyone changing that option will have to enable runtime datagen.

PR #513, opened 2023-06-12
Add null check in `PipeColorProvider#getColor`

### LINKED ISSUE #512: `PipeColorProvider` errors when world argument is `null`
The `BlockColor#getColor` method takes 4 arguments, 2 of these are marked `@Nullable` which includes the world (`BlockAndTintGetter`) argument. This can indeed be null, for example when called from inside the `BlockRenderDispatcher#renderSingleBlock` method which doesn't take a world argument.

The implementation of `BlockColor#getColor` in `PipeColorProvider` simply calls `BlockAndTintGetter#getBlockEntity` on the world argument without a null check. Whenever the world argument is `null`, this leads to a null pointer exception.
https://github.com/AztechMC/Modern-Industrialization/blob/d4b89ed39d9f5437bee9f24bc14e1c15b967c4c9/src/client/java/aztech/modern_industrialization/pipes/impl/PipeColorProvider.java#L34-L37

In my mod Entangled I happen to use `BlockRenderDispatcher#renderSingleBlock` to render other blocks in the world. When the rendered block is a pipe from Modern Industrialization, this leads to a crash.
Original issue here: https://github.com/SuperMartijn642/Entangled/issues/61

SuperMartijn642:
This pull request fixes #512 by adding a null check for both the world and position arguments in `PipeColorProvider#getColor`. If either argument is `null`, the method will simply return `-1`.
```diff
    @Override
    public int getColor(BlockState state, BlockAndTintGetter world, BlockPos pos, int tintIndex) {
-       if (world.getBlockEntity(pos) instanceof PipeBlockEntity) {
+       if (world != null && pos != null && world.getBlockEntity(pos) instanceof PipeBlockEntity) {
            ...
        }
        return -1;
    }
```

### DIFF
--- a/src/client/java/aztech/modern_industrialization/pipes/impl/PipeColorProvider.java
+++ b/src/client/java/aztech/modern_industrialization/pipes/impl/PipeColorProvider.java
@@ -33,7 +33,7 @@ import net.minecraft.world.level.block.state.BlockState;
 public class PipeColorProvider implements BlockColor {
     @Override
     public int getColor(BlockState state, BlockAndTintGetter world, BlockPos pos, int tintIndex) {
-        if (world.getBlockEntity(pos) instanceof PipeBlockEntity) {
+        if (world != null && pos != null && world.getBlockEntity(pos) instanceof PipeBlockEntity) {
             PipeBlockEntity entity = (PipeBlockEntity) world.getBlockEntity(pos);
             if (entity == null)
                 return -1;


Technici4n:
Thanks for the fix, this is fine to merge but will just not render the pipe with Entangled as far as I can tell. 

To render a generic block on Fabric, you should use a render method which will correctly allow access to the block entity data, i.e. one which takes a `BlockAndTintGetter`. This probably means creating a fake `BlockAndTintGetter` and passing that to the block renderer.

PR #518, opened 2023-06-19
Unhardcode cable tiers.

Fuyukai:
This converts ``CableTier`` into an interface, adds a default implementation via a record, and adjusts the code to not rely on enum behaviour. Obviously not finished yet. No clue if it actually works but hey, it compiles.

Points to discuss before merging:

- Adding tiers right now won't really work because transformers and the likes assume that EV<->Superconductor is a thing. 
- Making ``CableTier`` own a ``MutableComponent`` feels kinda gross.
- ``MachineCasings#casingFromCableTier`` needs to be adjusted to support this.
- Adding things like transformers, energy storage, and friends for the cable tier should probnably be exposed publically for addons/KubeJS.

### DIFF
--- a/src/main/java/aztech/modern_industrialization/api/energy/CableTier.java
+++ b/src/main/java/aztech/modern_industrialization/api/energy/CableTier.java
@@ -24,62 +24,64 @@
 package aztech.modern_industrialization.api.energy;
 
 import aztech.modern_industrialization.MIText;
-import net.minecraft.network.chat.Component;
+import java.util.Objects;
+import java.util.SortedSet;
+import java.util.TreeSet;
+import javax.annotation.Nullable;
+import net.minecraft.network.chat.MutableComponent;
 
-public enum CableTier {
-    LV("LV", "lv", 32, MIText.CableTierLV),
-    MV("MV", "mv", 32 * 4, MIText.CableTierMV),
-    HV("HV", "hv", 32 * 4 * 8, MIText.CableTierHV),
-    EV("EV", "ev", 32 * 4 * 8 * 8, MIText.CableTierEV),
-    SUPERCONDUCTOR("Superconductor", "superconductor", 128000000, MIText.CableTierSuperconductor);
+/**
+ * A single tier of cables.
+ */
+public interface CableTier extends Comparable<CableTier> {
+    SortedSet<CableTier> TIERS = new TreeSet<>();
 
-    public final String englishName;
-    public final String name;
-    public final long eu;
-
-    public final String translationKey;
-    public final Component englishNameComponent;
-
-    CableTier(String englishName, String name, long eu, MIText englishNameComponent) {
-        this.englishName = englishName;
-        this.name = name;
-        this.eu = eu;
-        this.translationKey = "text.modern_industrialization.cable_tier_" + name;
-        this.englishNameComponent = englishNameComponent.text();
-    }
-
-    public static CableTier getByName(String tier) {
-        for (CableTier cableTier : values()) {
-            if (cableTier.name.equals(tier)) {
-                return cableTier;
-            }
+    /**
+     * Gets a table tier by internal name.
+     *
+     * @param name The internal name of the tier, e.g. 'LV'.
+     * @return If it exists, the added cable tier; otherwise, <code>null</code>.
+     */
+    static @Nullable CableTier getByName(String name) {
+        for (var tier : TIERS) {
+            if (Objects.equals(tier.name(), name))
+                return tier;
         }
+
         return null;
     }
+
+    static CableTier create(String name, String englishName, long eu, MutableComponent text) {
+        return new CableTierImpl(name, englishName, eu, eu * 8, text);
+    }
+
+    static CableTier create(String name, String englishName, long eu, long maxTransfer, MutableComponent text) {
+        return new CableTierImpl(name, englishName, eu, maxTransfer, text);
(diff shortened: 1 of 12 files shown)


Technici4n:
Buildscript changes should go in a separate PR. (And you should get rid of jsr305.)

Technici4n:
The biggest question is of course what should happen with MI's own EV<->Superconductor transformers

Azercoco:
I think the only reasonable solution is juste that they will always exist and it's up to the modpack dev to desactivate them.

Fuyukai:
I separated out the gradle changes into its own PR (but rebased this one on top of it though).

> The biggest question is of course what should happen with MI's own EV<->Superconductor transformers

yeah I agree with the other comment, just jei hide + remove the recipe and register your own transformer for ZPM<->superconductor instead.

Technici4n:
I think a rebase would be good for reviewing now that the buildscript stuff was merged. :)

src/main/java/aztech/modern_industrialization/api/energy/CableTierImpl.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package aztech.modern_industrialization.api.energy;
+
+import net.minecraft.network.chat.MutableComponent;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Record-based implementation of a CableTier. Use the utility functions on the interface instead.
+ */
+record CableTierImpl(
```
Technici4n, line 32: I think this is a bit weird. We don't need an interface and a separate implementation, the whole thing can be a class most likely.

src/main/java/aztech/modern_industrialization/api/energy/CableTierImpl.java
```diff
@@ -0,0 +1,42 @@
...
+        String name,
+        String englishName,
+        long eu,
+        long maxTransfer,
+        MutableComponent englishTextComponent) implements CableTier {
```
Technici4n, line 37: I forgot if I already said so, but the solution here is to store a (non mutable) `Component` and then `copy()` it in the getter. This should just be a normal `CableTier` class I think, no need for a record.

src/main/java/aztech/modern_industrialization/api/energy/CableTier.java
```diff
@@ -24,62 +24,65 @@
...
-        return null;
+    static CableTier create(String name, String englishName, long eu, long maxTransfer, MutableComponent text) {
+        return new CableTierImpl(name, englishName, eu, maxTransfer, text);
     }
 
+    CableTier LV = create("lv", "LV", 32, MIText.CableTierLV.text());
+    CableTier MV = create("mv", "MV", 32 * 4, MIText.CableTierMV.text());
+    CableTier HV = create("hv", "HV", 32 * 4 * 8, MIText.CableTierHV.text());
+    CableTier EV = create("ev", "EV", 32 * 4 * 8 * 8, MIText.CableTierEV.text());
+    CableTier SUPERCONDUCTOR = create("superconductor", "SUPERCONDUCTOR", 128_000_000, MIText.CableTierSuperconductor.text());
```
Technici4n, line 67: Something else I'll ask too is that a KubeJS event be added too in this PR, to make sure that all the functionality is exposed from the start. I think some inspiration can be drawn from the way EBF tiers are registered.

src/main/java/aztech/modern_industrialization/machines/init/SingleBlockSpecialMachines.java
```diff
@@ -23,6 +23,8 @@
  */
 package aztech.modern_industrialization.machines.init;
 
+import static aztech.modern_industrialization.machines.init.MachineRegistrationHelper.*;
```
Technici4n, line 26: Please remove this, it's quite a large diff for no reason. :P

src/main/java/aztech/modern_industrialization/machines/blockentities/AbstractStorageMachineBlockEntity.java
```diff
@@ -91,10 +89,6 @@ public MIInventory getInventory() {
         return MIInventory.EMPTY;
     }
 
-    public static MachineCasing getCasingFromTier(CableTier from, CableTier to) {
```
Technici4n, line 94: I think the MachineCasing should be a property of the CableTier. MachineCasings are already easy to register in the KubeJS API so this won't be a problem.

src/main/java/aztech/modern_industrialization/machines/init/SingleBlockSpecialMachines.java
Technici4n: Though I must say it is quite nicer without the MachineRegistrationHelper, so this is fine actually.

Technici4n:
Btw if you could address the other stuff it would be nice. So I can release this before starting the 1.20 port.

Fuyukai:
Sorry, but other things came up now and I'm working on those mostly instead. I can rebase this onto 1.20 when you've done that instead.

Technici4n:
No worries! Thanks for letting me know :)

James103:
Does this PR allow modification of the five built-in cable tiers? If not, should it?
For example, making HV and EV be 512 and 2,048 EU/t instead of the default 1,024 and 8,192 EU/t, giving additional voltage space for IV and higher tiers.

PR #519, opened 2023-06-19
update buildscript + gradle to work on gradle 8.1.1

Fuyukai:
- update loom to 1.2.7
- apply some tape to some mods that use really ancient fabric loader versions (TR energy) that loom inexplicitly decided to break with 1.2.x
- bump wrapper version to 8.1.1, so now gradle can run under a java 19/20/etc host
- add toolchain config so that if youre using java 20 host you can avoid explosions
- switch out some ``javax.annotations`` for jetbrains ones. JB is used in the rest of the project and i assume newer loom stops putting JSR305 on compileClasspath for now.

also bumps some deps for good measure whilst i was there (except wthit because that has API changes that seem to be being tracked in a different PR.) fabric api also isn't bumped because it inexplicitly explodes with mixin errors.

separated out from #518.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,10 +1,10 @@
 plugins {
-    id 'fabric-loom' version '1.0-SNAPSHOT'
+    id 'fabric-loom' version '1.2.7'
     id 'maven-publish'
     id 'com.diffplug.spotless' version '5.14.0'
     id 'com.matthewprenger.cursegradle' version '1.4.0'
     id "com.modrinth.minotaur" version "2.+"
-    id 'io.github.juuxel.loom-quiltflower' version '1.8.0'
+    id 'io.github.juuxel.loom-quiltflower' version '1.10.0'
 }
 
 apply plugin: 'java-library'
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -268,6 +269,22 @@ dependencies {
     }
 }
 
+configurations {
+    // due to a change in newer loom, loader isn't remapped
+    // except some mods pull in an older version of loader that *is* remapped
+    // so it has to be excluded explicitly.
+
+    modRuntimeOnly.exclude group: "net.fabricmc", module: "fabric-loader"
```
Technici4n, line 277: Is loader still there when running the game with this?

build.gradle
```diff
@@ -276,10 +293,6 @@ processResources {
     }
 }
 
-loom {
-    splitEnvironmentSourceSets()
```
Technici4n, line 280: This was before `sourceSets` for a reason, this probably breaks stuff

build.gradle
```diff
@@ -392,7 +399,7 @@ import com.google.gson.GsonBuilder
 
 final class JsonOrderStep {
     private JsonOrderStep() {}
-    private static final Gson GSON = new GsonBuilder().disableHtmlEscaping().setLenient().setPrettyPrinting().create();
+    private static final Gson GSON = new GsonBuilder().disableHtmlEscaping().setLenient().setPrettyPrinting().create()
```
Technici4n, line 402: Why do you remove the semicolons? :(

build.gradle
Fuyukai: yes, this just excludes it transitively. loader is in the modImplementation configuration, not modRuntimeOnly.

build.gradle
Fuyukai: works fine in my IDE. ``net.minecraft.client`` isn't available from main sourceSet, but it is available in ``client`` sourceset.

build.gradle
Fuyukai: intellij alt-enter has its own mind. it's groovy it doeesnt even need semicolons!

PR #566, opened 2023-09-03
Forge Hammer noises

TheStaticVoid:
I made this suggestion in [Discord](https://discord.com/channels/754461970870173767/1138287541213216809) awhile ago, but it seemed like a pretty quick addition.

### DIFF
--- a/src/main/java/aztech/modern_industrialization/blocks/forgehammer/ForgeHammerScreenHandler.java
+++ b/src/main/java/aztech/modern_industrialization/blocks/forgehammer/ForgeHammerScreenHandler.java
@@ -59,6 +59,7 @@ public class ForgeHammerScreenHandler extends AbstractContainerMenu {
     private final ContainerLevelAccess context;
     private final Level world;
     private final Player player;
+    private long lastSoundTime = 0;
 
     private ItemStack inputStackCache = ItemStack.EMPTY, toolStackCache = ItemStack.EMPTY;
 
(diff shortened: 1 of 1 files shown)


src/main/java/aztech/modern_industrialization/blocks/forgehammer/ForgeHammerScreenHandler.java
```diff
@@ -114,6 +115,14 @@ public boolean mayPlace(ItemStack stack) {
             @Override
             public void onTake(Player player, ItemStack stack) {
                 ForgeHammerScreenHandler.this.onCraft();
+                // Don't play the sound multiple times within the same tick
+                // Prevents the sound being played a lot when shift-clicking the output into your inventory
+                context.execute((world, pos) -> {
+                    if (lastSoundTime < world.getGameTime()) {
+                        world.playSound(null, pos, SoundEvents.SMITHING_TABLE_USE, SoundSource.BLOCKS, 1.0F, 1.0F);
```
Technici4n, line 122: If this is called both on the client and the server, the `null` should probably be replaced by `player`? Can't check the vanilla code right now... 😉

TheStaticVoid: If you were to provide the player it would play the sound twice I believe. Once on the server and on the player, by providing `null` it only plays on the server side. 

On the client side it won't do anything, since player is null it will fail the `if (player == this.minecraft.player)` in `ClientLevel`

TheStaticVoid: Also to note, this is the same way the tool break sound effect is played in `onCraft` in this same class [here](https://github.com/AztechMC/Modern-Industrialization/blob/master/src/main/java/aztech/modern_industrialization/blocks/forgehammer/ForgeHammerScreenHandler.java#L228)

TheStaticVoid:
The original suggestion was for smithing table sounds. I find the anvil noise to be extremely annoying, but Mojang got the smithing table just right.

PR #567, opened 2023-09-04
Fix yarn mappings conflict.

### LINKED ISSUE #520: Incompatibility with yarn development environments
Modern industrialization is currently incompatible with mod development environments using yarn due to a remapping conflict. It seems like the fix is to simply rename a single method. I encountered this while investigating https://github.com/MattiDragon/ExtendedDrawers/issues/66 that might be related to this mod.

[`HackySlot`](https://github.com/AztechMC/Modern-Industrialization/blob/master/src/main/java/aztech/modern_industrialization/inventory/HackySlot.java#L43-L45) has methods called `getStack` and `setStack`. The issue arises when an attempt is made to remap modern industrialization to yarn because the yarn names for `getItem` and `set` are `getStack` and `setStack` respectively. So, because the signatures match exactly we'd end up with two of the same method. This causes fabric loader to abort the remapping as it should. Renaming the method should fix this.

Similar to https://github.com/Technici4n/Modern-Dynamics/issues/73

Fuyukai:
Just renames getStack() to getRealStack. Closes #520.

### DIFF
--- a/src/main/java/aztech/modern_industrialization/inventory/ConfigurableItemStack.java
+++ b/src/main/java/aztech/modern_industrialization/inventory/ConfigurableItemStack.java
@@ -219,12 +219,12 @@ public class ConfigurableItemStack extends AbstractConfigurableStack<Item, ItemV
         }
 
         @Override
-        protected ItemStack getStack() {
+        protected ItemStack getRealStack() {
             return key.toStack((int) amount);
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


Technici4n:
Do you mind targeting `master`? We'll work on 1.19 and merge forward to 1.20 for the time being

Fuyukai:
Sure, I'll close this and cherrypick it over quickly.

PR #573, opened 2023-09-14
Kind of stupid quantum sword improvements

Despacito696969:
1. Quantum sword is no longer affected by damage attribute. Previously using weakness would disable (non-enchanted) quantum sword. This makes witches no longer overpowered.
2. Mobs holding quantum sword will now do the correct amount of damage to their victims. I'm not sure who would give a quantum sword to a mob. I would never do that.


James103:
What death message do you get when a mob wielding a quantum sword hits you?
Expected result (for a zombie): "Player was slain by Zombie" (replace "Player" with your in-game name)

Despacito696969:
> What death message do you get when a mob wielding a quantum sword hits you? Expected result (for a zombie): "Player was slain by Zombie" (replace "Player" with your in-game name)

I thought it's "didn't want to live in the same world as Zombie" (the same as if you would kill an animal), but it's actually "fell out of the world", I forgot to check it after some later changes. It's an easy fix, but while testing I also found out that in normal MI with QS player kill can get two different messages, because MI first damages a mob with item, like usual and then it kills it, which results with one message if the hurt killed a mob "was slain by" or player and if not with the "didn't want to live in the same world" message. I did a fix that makes everything "was slain by" but I don't know, "didn't want to live in the same world" just sounds cooler.

Technici4n (changes requested):
Not sure what these mixins do (no computer atm), but I don't want any additional mixin.

Despacito696969:
Couldn't do the mob thingy without mixins, I don't know how to do that or if it's possible. Also made quantum sword not destroy blocks in creative, just like vanilla swords.

Despacito696969:
Also, now quantum sword can kill people with shields and people with creative mode, I think that's an improvement.

PR #593, opened 2023-10-30
Modular tools

JustinHuPrime:
Here's my attempt at a modular tools system.

The base item for a modular tool is a casing (in one of four tiers). Tool components consist of an energy converter (motor/pump), energy storage (battery/tank), and tool head (rotary blade or drill head). Tools also can have modules; these apply enchantment effects. Tools are modified at a tool station block, similar to the forge hammer. See https://github.com/AztechMC/Modern-Industrialization/blob/4f278b3717336ea17c364ddd53c0dcf4b67f3d33/docs/MODULAR_TOOLS.md for more details.

**Note that battery capacity in portable storage units have been reduced so a battery has one and only one capacity, instead of one capacity in a PSU and one in a modular tool.** I don't think this affects balance much. I might also consider a followup PR to reduce battery capacity in PSUs to 6 or so.

Open issues:

- Tools lose their stored energy/fuel when modified at a tool station: I think this could be solved by putting the fuel storage in the tank component and avoiding clearing the energy storage unless the tool's had its energy storage module changed out.
- Textures are very much programmer art: I don't think I've got the skills to solve this.
- Playtesting is probably needed to find any edge-case bugs: I'm willing to commit to supporting this part of MI
- Mutually exclusive enchantments can be applied (e.g. Fortune and Silk Touch); there's probably some special casing needed in the module slot code to solve this. (This can happen in Vanilla via the `/enchant` command, so the game handles the conflict okay.) I guess I could put a check in the tool station to prevent conflicting modules like I currently prevent duplicate modules.

Bugs to fix before merging:

- [x] Sweeping edge doesn't seem to work - do I need to tag the tool as a sword or something?
- [x] Attacking whilst the tool is drained of energy gives it negative energy
- [x] Tool mining speed is inconsistent when drained of energy - it should be universally zero
- [x] Enchantments are effective while tool is drained of energy


JustinHuPrime:
Build seems to be failing for spotless violations around the license header - do you want me to apply the outdated and slightly incorrect header anyways?

JustinHuPrime:
Regarding Sweeping Edge not working, I think I'd need to add the enchantment NBT tag dynamically; I guess I'd do this when the tool gets a rebuild?

JustinHuPrime:
Aha - sweeping edge only works if the held item is a SwordItem

Azercoco:
I will be testing this. Be sure you correctly apply `gradlew spotlessApply` so you don't introduce changes in the formatting.

JustinHuPrime:
I'm aware spotless is failing - it's because I'm using a different license header. I think the one hardcoded for use is incorrect.

Azercoco:
I already found a bug : doing shift-click in the tool stations allows to put more than one battery or motor.

Technici4n:
Please use the hardcoded header, it is good enough. The year is just an annoyance and not really relevant for what we do here. (I doubt we'll get to 80 years after the death of both authors). And technically it should mention "and contributors" but in practice it doesn't matter either.

JustinHuPrime:
> I already found a bug : doing shift-click in the tool stations allows to put more than one battery or motor.

Oops, how embarrassing. Fixed.

> Please use the hardcoded header, it is good enough. The year is just an annoyance and not really relevant for what we do here. (I doubt we'll get to 80 years after the death of both authors). And technically it should mention "and contributors" but in practice it doesn't matter either.

Fixed.

Technici4n:
Closing this PR since it went stale. I still like the idea of modular tools but I am not sure that they are the best fit for MI.

PR #605, opened 2023-12-14
KubeJS support for the Nuclear Reactor and Fusion Reactor

globalista-dev:
**Todo list**
- [x] Custom fuel rods
- [x] Custom control rods
- [x] Control rod part 
- [x] Custom isotope params
- [x] Custom isotope fuel params
- [x] Add documentation for custom fuel rods and control rods
- [x] Custom fluid nuclear components

Documentation needs further explanations


docs/ADDING_NUCLEAR_FUELS.md
```diff
@@ -0,0 +1,185 @@
...
+
+Isotope Fuel Parameters and Isotope Parameters can also be added. They need a startup script and the events in `NuclearConstantEvents`
+
+## Add a fuel rod
+Here is an example script that adds a new `LE Americium MOX` material that has fuel rods.
+For a full tutorial on how to add new materials, refer to `ADDING_MATERIALS.md`
+
+```javascript
+MIMaterialEvents.addMaterials(event => {
+
```
Technici4n, line 13: I don't like these empty lines after an opening brace.

docs/ADDING_NUCLEAR_FUELS.md
```diff
@@ -0,0 +1,185 @@
...
+});
+```
+
+You can also mix two different isotope fuel parameters. This can be used to simulate MOX and MOX-type fuels.
+The resulting parameters will be `r * a + (1 - b) * b`, where `r` is the factor, `a` is the first set of parameters and `b` is the second 
+The example below illustrates how a fuel rod can be created from the U238 and Pu239 parameters
+This generates the same parameters as LE MOX fuel
+
+``` javascript
+[...]
```
Technici4n, line 38: You should provide full examples to reduce the amount of questions by people not familiar with JS.

src/main/java/aztech/modern_industrialization/compat/kubejs/nuclear/CreateIsotopeFuelParamsEventJS.java
```diff
@@ -0,0 +1,58 @@
...
+import aztech.modern_industrialization.nuclear.NuclearConstant;
+import dev.latvian.mods.kubejs.event.EventJS;
+import java.util.SortedMap;
+import java.util.TreeMap;
+
+public class CreateIsotopeFuelParamsEventJS extends EventJS {
+
+    public static SortedMap<String, NuclearConstant.IsotopeFuelParams> FUEL_PARAMS = new TreeMap<>();
+
+    public void createIsotopeFuelParams(String name, double thermalAbsorbProba, double thermalScatterings, int maxTemp, int tempLimitLow,
```
Technici4n, line 35: I think that these events are a bit of a roundabout way to go about things. My intuition is that the registration of nuclear components can probably be simplified a little to make this unnecessary? Not sure

src/main/java/aztech/modern_industrialization/nuclear/FluidNuclearComponents.java
```diff
@@ -51,49 +56,51 @@ public class FluidNuclearComponents {
...
         Fluid fluid = variant.getFluid();
 
-        if (fluid == Fluids.WATER) {
-            return WATER;
-        } else if (fluid == MIFluids.HEAVY_WATER.asFluid()) {
-            return HEAVY_WATER;
-        } else if (fluid == MIFluids.HIGH_PRESSURE_WATER.asFluid()) {
-            return HIGH_PRESSURE_WATER;
-        } else if (fluid == MIFluids.HIGH_PRESSURE_HEAVY_WATER.asFluid()) {
-            return HIGH_PRESSURE_HEAVY_WATER;
```
Technici4n, line 61: This whole function can be a `get` call at this point. :P

Technici4n (changes requested):
very promising, just a bit roundabout (but that's the fault of the current system, not your PR)

src/main/java/aztech/modern_industrialization/compat/kubejs/nuclear/CreateIsotopeFuelParamsEventJS.java
globalista-dev: Yeah, I thought about that as well. I couldn't find a way to make it less complicated, but I'll try and come up with a better way to do things

globalista-dev:
> I don't like these empty lines after an opening brace.
Fixed

> You should provide full examples to reduce the amount of questions by people not familiar with JS.
Added full examples to everything

src/main/java/aztech/modern_industrialization/nuclear/FluidNuclearComponents.java
globalista-dev: The _of_ function?

Technici4n: yeah - `get` will return null if the value is not in the map already

globalista-dev: You're right xD
Fixed it

globalista-dev:
Moved both params to the same class (RegisterNuclearParams) and to the MIRegistrationKubeJSEvents interface. Both types of params are using the same map, but that shouldn't really be a problem.

Technici4n (approved):

PR #616, opened 2024-01-31
Addition of Japanese translation

konumatakaki:
Translated into Japanese. Added files and folders are as follows:
`src/main/resources/assets/modern_industrialization/lang/ja_jp`
`src/main/resources/assets/modern_industrialization/patchouli_books/book/ja_jp/`

My translation should be complete.


konumatakaki:
Looks like there was a duplicate file, I'll fix it.

PR #710, opened 2024-05-07
Add rods to the tagged parts list for more mod compatibility

cplir-c:
I have checked that this builds and it looks like it should work --
Is there more testing I should do?


Technici4n:
This is a good start. You also need to go through every recipe that uses rods in `src/main/resources` and change that to the tag.

Technici4n:
Made a few edits and now we're good to go. Thank you. :)

PR #728, opened 2024-05-29
Add machine casing voltage tooltip

Swedz:
A couple things I should mention:

I wasn't able to get the datagen to run properly, so I figured that could be run at a later time before a release is published.

Also, the code in the tooltip is the exact same as the code in CasingComponent#getCasingTier, but I wasn't sure if I should reference that since its a method made for the component. It may be worthwhile moving this getCasingTier method elsewhere and using that here (or anywhere else that may need to reference cable tier casings).

### DIFF
--- a/src/main/java/aztech/modern_industrialization/MIText.java
+++ b/src/main/java/aztech/modern_industrialization/MIText.java
@@ -145,6 +145,7 @@ public enum MIText {
     LockingModeOff("Lock editing disabled"),
     LockingModeOn("Lock editing enabled"),
     LubricantTooltip("Right-Click on Electric Machine : consume %s mb for 1 efficiency tick"),
+    MachineCasingVoltage("Allows machines to accept %s power"),
     MachineUpgrade("Electric Machine Upgrade : Max Overclock +%s"),
     MachineUpgradeStack("Total Stack Upgrade +%s"),
     MaxEuProduction("Can produce up to %s"),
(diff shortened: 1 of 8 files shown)


src/main/java/aztech/modern_industrialization/MIText.java
```diff
@@ -248,7 +248,8 @@ public enum MIText {
     Unlocked("Unlocked"),
     UseItemToChange("Right-click to change."),
     WaterPercent("Water: %s %%"),
-    Whitelist("Whitelist mode enabled");
+    Whitelist("Whitelist mode enabled"),
+    MachineCasingVoltage("Allows machines to accept %s power");
```
Technici4n: This enum is sorted alphabetically. :P

Swedz: oh lol, i didnt realize. ill fix that

src/main/java/aztech/modern_industrialization/MITooltips.java
```diff
@@ -386,6 +387,17 @@ public Component parse(Double ratio) {
             MIText.ConfigCardHelpItems5,
             MIText.ConfigCardHelpClear);
 
+    public static final TooltipAttachment MACHINE_CASING_VOLTAGE = TooltipAttachment.of(
+            (itemStack, item) -> {
+                var itemKey = BuiltInRegistries.ITEM.getKey(item);
```
Technici4n: This should use the static method in CasingComponent.

Swedz: sweet, will do

Technici4n:
You have format violations. You can run `gradlew spotlessApply` to fix them.

PR #732, opened 2024-06-05
Recipe Condition Icons

Swedz:
A couple things I should mention:
- I had to bump REI's version as well as NeoForge's to get it to run. I did not include these changes in the PR.
- The CustomProcessCondition codec seems broken. When joining a world that has a recipe with a custom condition it kicks the client and has an error in the log: `Failed to encode: Unknown machine process condition codec`. I'm not sure if this is related to my changes or if process conditions are just broken right now


Swedz:
This pr needed some work, I'll look into this again at some point soon

PR #797, opened 2024-07-23
Very updated ru_ru

Dorfyt:
Most of russian localisation looks like it is from very bad translator( example: chemical reactor achievment is translated as "Walter approves white"), so i approved ru_ru massively (though some strings still need to be updated)


Technici4n:
Hmm, would it be possible to apply that on crowdin instead? (https://crowdin.com/project/modern-industrialization) That is where translations are now managed.

Dorfyt:
Uploaded on crowdin

Technici4n:
Thank you! I pulled that in for 2.2.13-beta. :)

PR #838, opened 2024-08-20
Use namespaces

Swedz:
These are the changes we talked about a while ago, finally got around to rebasing them and getting spotless to work

[Discord convo](https://discord.com/channels/754461970870173767/760125612126896196/1271571517611708616)


Technici4n:
Thanks! I pushed a commit that removes the bridge helper methods with a single caller. Also found a use for `Util.makeDescriptionId`. :wink:

PR #846, opened 2024-08-24
Allow energy cables to be quickly replaced while holding wrench in of…

DarkVen0m:
Right now, it's very frustrating dealing with cable replacement when you have to change energy cables tier in whole setup.
You have to wrench every energy cable without touching other ones (fluids, items), which is kind of pixel-hunting with their current collision boxes. And only after that can you place new ones.
To avoid breaking current behaviour, when multiple energy cables can occupy the same block, replacement is only happening while you hold the wrench in offhand.
https://www.youtube.com/watch?v=iOVB5uSiE1Y

### DIFF
--- a/src/main/java/aztech/modern_industrialization/pipes/impl/PipeItem.java
+++ b/src/main/java/aztech/modern_industrialization/pipes/impl/PipeItem.java
@@ -134,19 +134,21 @@ public class PipeItem extends Item {
         BlockEntity be = world.getBlockEntity(pos);
         if (be instanceof PipeBlockEntity pipeBe) {
             var player = context.getPlayer();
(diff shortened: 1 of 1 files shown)


Technici4n:
I think that this feature makes sense in some form, but I am not sure that this is how I want it to be. Will think more about it.

DarkVen0m:
Another thing I thought of was ability to disable rendering of all types of pipes except some specific one - so another feature that was part of EIO pipes system.
But here it doesn't solve issue, replacement still would be long and tedious process.
So the only other way I find solving the issue - some tool to replace cables in the whole system so it can also keep connections from the previous one (unlike the feature in this PR).

PR #855, opened 2024-08-29
Fix Steam Drill mining or highlighting blocks it shouldn't & improve direction accuracy

### LINKED ISSUE #854: Steam Drill 3x3 has odd edge cases with rendering and breaking
1. When looking at a non-collidable block (like a sapling) and there's a solid block through it where your cursor is, it will render the highlight around the solid blocks instead of only just the non-collidable block.
<br>When you are in this state and begin to break the block, even though the block has a destroy speed of 0 (again, for a sapling) it doesn't treat it like that and will break the block using the mining speed of the highlighted block.
<br>Example:

2. When looking at a block that is not mineable (like bedrock), the mineable blocks around it are highlighted but not the targeted block. It would be more intuitive for it to only highlight the unmineable block so it doesn't seem like those other blocks can be mined by mining this unmineable block.
<br>Also related and more critical: when targeting an unmineable block (like bedrock) and mineable blocks are in the 3x3 mining area, the drill can break the bedrock directly.
<br>Example (it can be kind of hard to see, you have to look closely):

Swedz:
Fixes #854 & makes 3x3 mining direction more accurate (especially when on a remote server)

### DIFF
--- a/src/main/java/aztech/modern_industrialization/items/SteamDrillItem.java
+++ b/src/main/java/aztech/modern_industrialization/items/SteamDrillItem.java
@@ -34,6 +34,7 @@ import aztech.modern_industrialization.util.TextHelper;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.WeakHashMap;
 import java.util.function.BiConsumer;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
@@ -81,6 +82,7 @@ import net.neoforged.bus.api.EventPriority;
 import net.neoforged.neoforge.capabilities.Capabilities;
 import net.neoforged.neoforge.common.CommonHooks;
 import net.neoforged.neoforge.common.NeoForge;
+import net.neoforged.neoforge.event.entity.player.PlayerInteractEvent;
 import net.neoforged.neoforge.event.level.BlockDropsEvent;
 import net.neoforged.neoforge.fluids.FluidType;
 import org.apache.commons.lang3.mutable.Mutable;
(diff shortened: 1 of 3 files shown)


src/main/java/aztech/modern_industrialization/items/SteamDrillItem.java
```diff
@@ -260,24 +284,41 @@ private static void mergeDrops(BlockDropsEvent event) {
         event.getDrops().clear();
     }
 
+    private static void trackClickedFace(PlayerInteractEvent.LeftClickBlock event) {
+        Player player = event.getEntity();
+        if (!player.level().isClientSide()) {
+            PlayerInteractEvent.LeftClickBlock.Action action = event.getAction();
+            if (action == PlayerInteractEvent.LeftClickBlock.Action.START ||
+                    action == PlayerInteractEvent.LeftClickBlock.Action.STOP) {
+                lastClickedBlock.put(player, new ClickedBlock(event.getPos(), event.getFace()));
```
Technici4n:
```suggestion
                lastClickedBlock.put(player, new ClickedBlock(event.getPos().immutable(), event.getFace()));
```

Technici4n (changes requested):

src/main/java/aztech/modern_industrialization/items/SteamDrillItem.java
Technici4n: Probably doesn't matter here, but it's good practice when storing a BlockPos

Swedz: Good point, I missed this

Technici4n (approved):

PR #856, opened 2024-08-30
Equipment Quality of Life Improvements

Swedz:
- Reworked and cleaned up keybinds, its similar to the new system I wrote for EI which I liked

  - Proper translation for keybind category (before it just appeared as `modern_industrialization`)

- Moved and renamed `ActivatableChestItem` to `ActivatableItem` since it now applies to the steam drill as well

- Add 3x3 toggling by keybind for steam drill

- Drops from 3x3 mining in the steam drill are added directly to the inventory of the player and dropped on the ground if there's no space for it

- "Tool configuration:" header on the tool info sections to keep it separated from the tool settings

- Make steam drill tooltips more consistent with other item tooltips

  - 3x3 toggling keybind is shown in the help tooltip

  - Instead of displaying the silk touch enchantment, show a nice activated/deactivated line

  - Uses dashed line prefixes instead of numbered prefixes for the help tooltip

- Add tooltips for diesel tools to be more consistent with other item tooltips

  - Instead of displaying the active enchantment, show a nice mining mode line
  
  - Help tooltips similar to the steam drill ones to give some information on how the diesel tools can be fueled and used

- Use builtin NeoForge creative flight attribute for gravichestplate and quantum chestplate

- Running datagen for rei categories translation entries was broken by a previous pr (#838) so I quickly fixed that here too so that I could run datagen properly

- Removed space between semicolon and preceding character for all translations


src/main/java/aztech/modern_industrialization/items/armor/GraviChestPlateItem.java
```diff
@@ -48,14 +54,16 @@ public GraviChestPlateItem(Properties settings) {
...
+            return ItemAttributeModifiers.builder()
+                    .add(
+                            NeoForgeMod.CREATIVE_FLIGHT,
+                            new AttributeModifier(MI.id("gravichestplate_flight"), 1, AttributeModifier.Operation.ADD_VALUE),
+                            EquipmentSlotGroup.CHEST)
+                    .build()
+                    .withTooltip(false);
         }
-        return modifiers;
+        return ItemAttributeModifiers.EMPTY;
```
Swedz, line 66: I should point out that with this, modpacks or commands that add attributes to the gravichestplate will either A) not get shown or B) cause the "When on Body:" tooltip show twice. Not sure how much of a concern this may be, though.

Swedz: Same goes for the quantum armor, but that was always the case.

src/main/java/aztech/modern_industrialization/items/SteamDrillItem.java
```diff
@@ -290,7 +291,7 @@ public boolean mineBlock(ItemStack stack, Level world, BlockState state, BlockPo
             }
         });
         totalDrops.forEach(itemStack -> {
-            Block.popResource(world, miner.blockPosition(), itemStack);
+            ItemHandlerHelper.giveItemToPlayer(p, itemStack);
```
Swedz, line 340: This has incompatibility with the Bank Storage mod, where it doesn't pick up the items mined by the drill in 3x3 mode. I'm talking with the dev to see if we can either get a workaround or a fix for this. Just something to keep in mind :)

Swedz: Worked it out with them and they said they'll publish a fix in <1 week

src/client/java/aztech/modern_industrialization/MIKeybinds.java
```diff
@@ -0,0 +1,106 @@
...
+
+    public static void init(RegisterKeyMappingsEvent event) {
+        MAPPINGS.forEach((m) -> event.register(m.holder().get()));
+    }
+
+    public static final String CATEGORY = Util.makeDescriptionId("key.categories", MI.id(MI.ID));
+
+    public static final Keybind TOGGLE_FLIGHT = create(
+            "toggle_flight",
+            (id) -> new KeyMapping(
```
Technici4n: This should pass `KeyConflictContext.IN_GAME` to avoid conflicts with GUI-only keybindings.

src/client/java/aztech/modern_industrialization/MIKeybinds.java
```diff
@@ -0,0 +1,106 @@
...
+
+public class MIKeybinds {
+    private static final Set<Keybind> MAPPINGS = Sets.newHashSet();
+
+    public static Set<Keybind> getMappings() {
+        return Collections.unmodifiableSet(MAPPINGS);
+    }
+
+    public static void init(RegisterKeyMappingsEvent event) {
+        MAPPINGS.forEach((m) -> event.register(m.holder().get()));
```
Technici4n: I would generally remove the parentheses and write `m ->`. Feel free to fix or not, it's not important. (There are other places too).

src/main/java/aztech/modern_industrialization/mixin/ItemStackMixin.java
```diff
@@ -0,0 +1,51 @@
...
+ * {@link ItemAttributeModifiers#withTooltip(boolean)} (<code>false</code> as the input) and the item doesn't have an attribute component, it will
+ * still display the defaulted attribute. This is because {@link ItemStack#addAttributeTooltips(Consumer, Player)} directly uses
+ * {@link ItemStack#getOrDefault(DataComponentType, Object)} to get the item attribute modifiers rather than
+ * {@link IItemStackExtension#getAttributeModifiers()}. Despite the comment in the code saying otherwise, this is the cause of the issue.
+ */
+@Mixin(ItemStack.class)
+public class ItemStackMixin {
+    @WrapOperation(method = "addAttributeTooltips", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/item/ItemStack;getOrDefault(Lnet/minecraft/core/component/DataComponentType;Ljava/lang/Object;)Ljava/lang/Object;"))
+    private Object addAttributeTooltips(ItemStack instance, DataComponentType dataComponentType, Object defaultValue, Operation<Object> original) {
+        return instance.getAttributeModifiers();
```
Technici4n: This mixin might break other mods (who knows). It's better to limit the redirection to the relevant MI items, rather than applying a generic patch.

Technici4n (changes requested):

src/main/java/aztech/modern_industrialization/network/armor/ActivateItemPacket.java
```diff
@@ -23,28 +23,31 @@
...
-public record ActivateChestPacket(boolean activated) implements BasePacket {
-    public static final StreamCodec<ByteBuf, ActivateChestPacket> STREAM_CODEC = ByteBufCodecs.BOOL
-            .map(ActivateChestPacket::new, ActivateChestPacket::activated);
+public record ActivateItemPacket(EquipmentSlot slot, boolean activated) implements BasePacket {
+    public static final StreamCodec<ByteBuf, ActivateItemPacket> STREAM_CODEC = ByteBufCodecs
+            .fromCodec(RecordCodecBuilder.create((instance) -> instance
+                    .group(
+                            EquipmentSlot.CODEC.fieldOf("slot").forGetter(ActivateItemPacket::slot),
+                            Codec.BOOL.fieldOf("activated").forGetter(ActivateItemPacket::activated))
+                    .apply(instance, ActivateItemPacket::new)));
```
Natanaelel: I happened to stumble upon this. Please use `StreamCodec.composite`.

src/client/java/aztech/modern_industrialization/MIKeybinds.java
Swedz: There's a few places throughout MI that does that, and it's a personal preference of mine. I reverted the ones I did that I noticed. If there's others I missed let me know

src/main/java/aztech/modern_industrialization/network/armor/ActivateItemPacket.java
```diff
@@ -23,28 +23,29 @@
...
 import net.minecraft.world.entity.EquipmentSlot;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.ItemStack;
 
-public record ActivateChestPacket(boolean activated) implements BasePacket {
-    public static final StreamCodec<ByteBuf, ActivateChestPacket> STREAM_CODEC = ByteBufCodecs.BOOL
-            .map(ActivateChestPacket::new, ActivateChestPacket::activated);
+public record ActivateItemPacket(EquipmentSlot slot, boolean activated) implements BasePacket {
+    public static final StreamCodec<RegistryFriendlyByteBuf, ActivateItemPacket> STREAM_CODEC = StreamCodec.composite(
+            ByteBufCodecs.fromCodec(EquipmentSlot.CODEC),
```
Technici4n: You can use `NeoForgeStreamCodecs.enumCodec(EquipmentSlot.class)` I think.

Technici4n (approved):

PR #889, opened 2024-09-15
Updating the Japanese translation of the guidebook

konumatakaki:
Reflects updates as far as we have been able to ascertain.


Technici4n:
Thank you!

PR #893, opened 2024-09-17
Fix #817: Simplify internal recipe building

### LINKED ISSUE #817: Clean up buildRecipes in MachineCategory
Right now the `buildRecipes` method in `MachineCategory` builds the recipes for the furnace, cutting machine, and centrifuge manually instead of calling their respective `getRecipes(Level)` method from `MachineRecipeType`. Changing this to be more generic would make adding custom `MachineRecipeType`s with runtime generated recipes much easier :)

[Discord conversation about this for further context](https://discord.com/channels/754461970870173767/760125612126896196/1269211804500889622)

Swedz:
Closes #817 

This is the code I've been using in a mixin in Tesseract for a while now, and it seems to work properly. It doesn't feel quite optimized to me though.

### DIFF
--- a/src/client/java/aztech/modern_industrialization/compat/viewer/usage/MachineCategory.java
+++ b/src/client/java/aztech/modern_industrialization/compat/viewer/usage/MachineCategory.java
@@ -38,15 +38,14 @@ import aztech.modern_industrialization.machines.guicomponents.ProgressBarClient;
 import aztech.modern_industrialization.machines.init.MIMachineRecipeTypes;
 import aztech.modern_industrialization.machines.init.MachineTier;
 import aztech.modern_industrialization.machines.recipe.MachineRecipe;
-import aztech.modern_industrialization.machines.recipe.RecipeConversions;
 import aztech.modern_industrialization.thirdparty.fabrictransfer.api.fluid.FluidVariant;
 import aztech.modern_industrialization.util.TextHelper;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
-import java.util.Objects;
 import java.util.function.Consumer;
 import net.minecraft.ChatFormatting;
+import net.minecraft.client.Minecraft;
 import net.minecraft.core.RegistryAccess;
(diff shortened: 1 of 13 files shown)


Swedz:
It might be worth considering changing `ProxyableMachineRecipeType#fillRecipeList` to have the `RecipeManager` and `RegistryAccess` as parameters instead of the `Level`. I can make this change too if that would be desirable

Technici4n:
All the weird cache invalidation should not be called on the client; it might lead to outdated recipes in JEI forever. Passing the recipe manager and registry access is fine.
Speaking of the invalidation, I think it's also time to remove it, and instead invalidate after a data pack (re)load using an event. This should remain server-exclusive behavior.

Swedz:
~~Is the cache invalidation something I did in this change? I don't see what would cause it to be invalidating.
If there's something somewhere where this is happening I can change it to use a datapack event listener.~~
I found it, the stuff with `UPDATE_INTERVAL` in `ProxyableMachineRecipeType`, right?

Swedz:
As for passing the recipe manager and registry access, I'm a little conflicted on it. Passing the level is very useful for me in Extended Industrialization, since it lets me access the `PotionBrewing` instance for the level and generate recipes using it.

Technici4n:
>  I found it, the stuff with UPDATE_INTERVAL in ProxyableMachineRecipeType, right?

Yes. Maybe just making a method without caching works.

> Passing the level is very useful for me in Extended Industrialization, since it lets me access the PotionBrewing instance for the level and generate recipes using it.

True.

Swedz:
Looking for what event(s) to use for invalidating the cache here. I see `OnDatapackSyncEvent` and `AddReloadListenerEvent` which seems to be server side only, and `RecipesUpdatedEvent` which seems to be client side only. Why do you say that this invalidation should only be done server side? Do the recipes generated by `fillRecipeList` get sent to the client? Or does the client also need to build these themselves?

Swedz:
Could be moved to #900

PR #896, opened 2024-09-23
Fix pipe help tooltip to include sneak key

Swedz:


Swedz:
It says the check failed but I think github just timed out again

Technici4n (approved):

PR #900, opened 2024-09-24
Various internal machine api improvements

Swedz:
- ~~Improves model API to allow so it is more extensible~~ (moved to #945)

- ~~Use `ResourceLocation` for machine casing ids. If no namespace is provided, it defaults to MI's namespace~~ (moved to #945)

- ~~Widen visibility and overridability of some methods and variables in~~ ~~`ShapeMatcher`~~ (moved to #937), ~~`CasingComponent`~~ (moved to #939), and ~~`MultiblockMachineBlockEntity`~~ (moved to #937)

- ~~Make the `CableTier` of `EnergyHatch` and `CasingComponent` accessible~~ (moved to #939)

- ~~Improves shape matcher API to allow for machines to define their own custom shape matcher and also standardizes the various methods triggered relating to multiblock shape matching (lots of duplicate code is removed)~~ (moved to #937)

- ~~Improve the internal storing of components such that they can be accessed and modified in the same way for both and includes some extra utility methods (I think I covered all reasonable use cases)~~ (moved to #938)

- ~~Allow for vertical orientations in `OrientationComponent`~~ (moved to #940)

- ~~Allow `BackgroundRenderedSlot` to override the atlas used to render them~~ (moved to #941)


Swedz:
~~It might make sense to also add the `CableTierHolder` interface to electric machines and get the tier from their casing component. That can easily be done. Doing it for multi blocks is a little ambiguous since a multi block electric machine can have multiple types of hatches on it. It could just use the highest tier, though. Would either of these changes be acceptable?~~ (moved to #939)

Swedz:
Is there a reason `icomponents` in `MachineBlockEntity` is private? I've run into some situations where I need to remove a component or iterate over all components on a machine. There's workarounds obviously, but it might be worth making that public in this pr.

PR #913, opened 2024-10-08
tr_tr.json Turkish Localization (Mod + Book)

RuyaSavascisi:


Technici4n:
Thanks for the PR. The translations are done via crowdin so I've uploaded your translations there. The guidebook translation I will keep.

PR #918, opened 2024-10-27
Add support for recipe conditions to have icons

Swedz:
`CustomProcessCondition#makeCodec(boolean)` has no uses and could be removed - I'm not sure of what its purpose may have been so I haven't removed it quite yet.

Also, `ViewerCategory$WidgetList#item(double, double, double, double, ItemLike)` no longer has any uses due to my use of `widgets.drawable(...)` in `MachineCategory`. This could also be removed, but I want to make sure before doing it.


Technici4n:
Nice! I removed both `makeCodec` and `WidgetList#item`. The latter was written before `GuiGraphics`, when item rendering was messy and best left to each recipe viewer.

PR #932, opened 2024-11-21
Add structure-like support for multiblock shapes

Swedz:
This is in a "somewhat done" state. It is **not** ready to be merged. This relies on some other pull requests, and this needs to go through some testing and extra refinement first.

My plans for things that need to be addressed:
- The documentation needs to be updated to explain how to use this system. Including screenshots would be ideal

- Include the block requirements for framed blocks in the recipe viewer

- Include tooltip when looking at a block in the hologram that requires other blocks to go inside it (like framed blocks) / or maybe flip through the framed block without nbt and the one with

- [MIStructureTemplateManager](https://github.com/AztechMC/Modern-Industrialization/blob/9f04d831e6479e3ccf8bf86df79b194c0d5aa39b/src/main/java/aztech/modern_industrialization/machines/multiblocks/structure/MIStructureTemplateManager.java) is pretty messy and I am unhappy with how it works. I'm not sure what could be done about this.

- [ShapeTemplate](https://github.com/AztechMC/Modern-Industrialization/blob/9f04d831e6479e3ccf8bf86df79b194c0d5aa39b/src/main/java/aztech/modern_industrialization/machines/multiblocks/ShapeTemplate.java) may benefit from a bit of a rewrite that makes it better follow OOP standards. Right now, when I need to check if a ShapeTemplate is a structure ShapeTemplate, I have to iterate over all of the members and ensure they are StructureMembers. Quite messy :P

Here are some things that may need addressing:
- This system at the moment is not used for the Large Tank and Nuclear Reactor multiblock shapes. Reason being both of these have functionality inherently linked to their shapes. No clue how this could be addressed, if at all. May be fine to just leave them as-is.

- While this system is being used for the Distillation Tower, there is a config option that lets you change the max height of the tower. As a quick "fix" for this, I have added structures for the Distillation Tower up to height 15 and it only uses them based on the config. The game will crash if the height is set to anything beyond 15, because the structure can't be found. Here are some possible solutions:
  - The distillation tower doesn't use the structure system.
  - The user is required to add structures for the distillation tower for higher height settings (how it works now).
  - Using something like datagen to create structure files and then having these taller tower structure files generate when enabling some config option.


James103:
For structures with block entity NBT requirements, if an NBT tag is not specified within the required NBT data but is present on the block entity being checked, does the check fail hard as the extra NBT must not be present, or is the extra NBT ignored?

For example, if I want to require a player head with a specific texture as part of the structure (with said head being obtainable by other means), will I need to specify all parts of the profile data, or can I leave out a part and still get a match when someone builds my custom multiblock?

Swedz:
It uses strict matching. If the NBT for a block in the file is completely omitted, then it will not check will not check against the NBT in world. If there is NBT present, it must be an exact match to the NBT in world. I have plans to add a structure wrench item to allow you to exclude specific blocks in the multiblock shape from checking NBT even if you have "Include NBT" enabled. I also have plans to allow for NBT being defined in the text boxes for the structure member block so you can specify what parts of the NBT to check against.

James103:
Is it possible to specify that the block at a specific position MUST be air lest the multiblock not form?
For example, I would like to add some kind of exhaust block to the structure of the Large Steam Boiler (and its advanced variants) and require the block above that to be air.

Swedz:
Yes, you can use the member block and set the member as air.

PR #937, opened 2024-11-25
Make machine shape matcher api more extensible

Swedz:
Contains some parts of #900

### DIFF
--- a/src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
+++ b/src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
@@ -33,6 +33,8 @@ import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.phys.BlockHitResult;
 
 public abstract class MultiblockMachineBlockEntity extends MachineBlockEntity {
+    protected ShapeMatcher shapeMatcher;
+
     public MultiblockMachineBlockEntity(BEP bep, MachineGuiParameters guiParams, OrientationComponent.Params orientationParams) {
         super(bep, guiParams, orientationParams);
         this.shapeValid = new ShapeValidComponent();
@@ -45,7 +47,61 @@ public abstract class MultiblockMachineBlockEntity extends MachineBlockEntity {
         return shapeValid.shapeValid;
     }
 
-    public abstract void unlink();
+    public ShapeMatcher getShapeMatcher() {
+        return shapeMatcher;
+    }
+
+    public ShapeMatcher createShapeMatcher() {
+        return new ShapeMatcher(level, worldPosition, orientation.facingDirection, getActiveShape());
+    }
+
+    protected void onLink() {
+    }
+
+    protected void onUnlink() {
+    }
+
+    protected void onRematch() {
+    }
+
+    protected void onMatchSuccessful() {
+    }
+
+    protected void onMatchFailure() {
+    }
+
+    protected void link() {
+        if (shapeMatcher == null) {
+            shapeMatcher = createShapeMatcher();
+            shapeMatcher.registerListeners(level);
+            onLink();
+        }
+        if (shapeMatcher.needsRematch()) {
+            shapeValid.shapeValid = false;
+            shapeMatcher.rematch(level);
+            onRematch();
+
+            if (shapeMatcher.isMatchSuccessful()) {
+                onMatchSuccessful();
+                shapeValid.shapeValid = true;
+            } else {
+                onMatchFailure();
+            }
+
+            if (shapeValid.update()) {
+                sync(false);
+            }
+        }
+    }
+
+    public void unlink() {
+        if (shapeMatcher != null) {
+            shapeMatcher.unlinkHatches();
+            shapeMatcher.unregisterListeners(level);
+            onUnlink();
+            shapeMatcher = null;
+        }
+    }
 
     @Override
     public boolean useWrench(Player player, InteractionHand hand, BlockHitResult hitResult) {
@@ -59,7 +115,7 @@ public abstract class MultiblockMachineBlockEntity extends MachineBlockEntity {
     }
 
     @Override
-    public final void setRemoved() {
+    public void setRemoved() {
         super.setRemoved();
         if (!level.isClientSide) {
             unlink();
--- a/src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
+++ b/src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
@@ -31,9 +31,11 @@ import aztech.modern_industrialization.machines.gui.MachineGuiParameters;
 import net.minecraft.world.InteractionHand;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.phys.BlockHitResult;
+import org.jetbrains.annotations.Nullable;
 
 public abstract class MultiblockMachineBlockEntity extends MachineBlockEntity {
-    protected ShapeMatcher shapeMatcher;
+    @Nullable
+    private ShapeMatcher shapeMatcher;
 
     public MultiblockMachineBlockEntity(BEP bep, MachineGuiParameters guiParams, OrientationComponent.Params orientationParams) {
         super(bep, guiParams, orientationParams);
@@ -47,45 +49,25 @@ public abstract class MultiblockMachineBlockEntity extends MachineBlockEntity {
         return shapeValid.shapeValid;
     }
 
-    public ShapeMatcher getShapeMatcher() {
-        return shapeMatcher;
-    }
-
     public ShapeMatcher createShapeMatcher() {
         return new ShapeMatcher(level, worldPosition, orientation.facingDirection, getActiveShape());
     }
 
-    protected void onLink() {
+    protected void onRematch(ShapeMatcher shapeMatcher) {
     }
 
-    protected void onUnlink() {
-    }
-
-    protected void onRematch() {
-    }
-
-    protected void onMatchSuccessful() {
-    }
-
-    protected void onMatchFailure() {
-    }
-
-    protected void link() {
+    protected final void link() {
         if (shapeMatcher == null) {
             shapeMatcher = createShapeMatcher();
             shapeMatcher.registerListeners(level);
-            onLink();
         }
         if (shapeMatcher.needsRematch()) {
             shapeValid.shapeValid = false;
             shapeMatcher.rematch(level);
-            onRematch();
+            onRematch(shapeMatcher);
 
             if (shapeMatcher.isMatchSuccessful()) {
-                onMatchSuccessful();
                 shapeValid.shapeValid = true;
-            } else {
-                onMatchFailure();
             }
 
             if (shapeValid.update()) {
@@ -94,11 +76,10 @@ public abstract class MultiblockMachineBlockEntity extends MachineBlockEntity {
         }
     }
 
-    public void unlink() {
+    public final void unlink() {
         if (shapeMatcher != null) {
             shapeMatcher.unlinkHatches();
             shapeMatcher.unregisterListeners(level);
-            onUnlink();
             shapeMatcher = null;
         }
     }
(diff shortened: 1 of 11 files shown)


src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
```diff
@@ -45,7 +47,61 @@ public boolean isShapeValid() {
...
+    }
+
+    protected void onMatchFailure() {
+    }
+
+    protected void link() {
+        if (shapeMatcher == null) {
+            shapeMatcher = createShapeMatcher();
+            shapeMatcher.registerListeners(level);
+            onLink();
```
Technici4n: Current MI doesn't need this. Do you?

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
```diff
@@ -45,7 +47,61 @@ public boolean isShapeValid() {
...
+        if (shapeMatcher.needsRematch()) {
+            shapeValid.shapeValid = false;
+            shapeMatcher.rematch(level);
+            onRematch();
+
+            if (shapeMatcher.isMatchSuccessful()) {
+                onMatchSuccessful();
+                shapeValid.shapeValid = true;
+            } else {
+                onMatchFailure();
```
Technici4n: Current MI doesn't need this. Do you?

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
```diff
@@ -45,7 +47,61 @@ public boolean isShapeValid() {
...
+    protected void onRematch() {
+    }
+
+    protected void onMatchSuccessful() {
+    }
+
+    protected void onMatchFailure() {
+    }
+
+    protected void link() {
```
Technici4n: Make it final.

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
```diff
@@ -45,7 +47,61 @@ public boolean isShapeValid() {
...
+                onMatchFailure();
+            }
+
+            if (shapeValid.update()) {
+                sync(false);
+            }
+        }
+    }
+
+    public void unlink() {
```
Technici4n: Make it final.

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
```diff
@@ -59,7 +115,7 @@ public boolean useWrench(Player player, InteractionHand hand, BlockHitResult hit
     }
 
     @Override
-    public final void setRemoved() {
+    public void setRemoved() {
```
Technici4n, line 99: Why not keep this final? When you make this kind of change it really needs to be explained. 😅

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
```diff
@@ -45,7 +47,61 @@ public boolean isShapeValid() {
...
+                sync(false);
+            }
+        }
+    }
+
+    public void unlink() {
+        if (shapeMatcher != null) {
+            shapeMatcher.unlinkHatches();
+            shapeMatcher.unregisterListeners(level);
+            onUnlink();
```
Technici4n: Current MI doesn't need this. Do you?

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
```diff
@@ -33,6 +33,8 @@
 import net.minecraft.world.phys.BlockHitResult;
 
 public abstract class MultiblockMachineBlockEntity extends MachineBlockEntity {
+    protected ShapeMatcher shapeMatcher;
```
Technici4n: Make this nullable and keep it private, IMO.

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
```diff
@@ -45,7 +47,61 @@ public boolean isShapeValid() {
         return shapeValid.shapeValid;
     }
 
-    public abstract void unlink();
+    public ShapeMatcher getShapeMatcher() {
```
Technici4n: Current MI doesn't need this. Do you?

src/main/java/aztech/modern_industrialization/machines/multiblocks/ShapeMatcher.java
```diff
@@ -47,14 +47,14 @@ public ShapeMatcher(Level world, BlockPos controllerPos, Direction controllerDir
         this.hatchFlags = toWorldPos(controllerPos, controllerDirection, template.hatchFlags);
     }
```
Technici4n, line 49: Why make all of this stuff `protected`?

Technici4n (changes requested):
I like the cleanup of `link` and `unlink`, it was long overdue.

Control flow in the multiblock code is already quite tricky, so make sure not to add any unnecessary virtual method, these really add mental overhead. Remember that complexity is the #1 enemy.

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
Swedz: Yes, I use it to use a custom shape matcher for my tesla tower to make it only form the multiblock if all energy hatches are the same tier to prevent converting energy tiers when transmitting wirelessly

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
Swedz: Yeah, I use it for updating the network for a tesla tower when the multiblock forms or fails to form

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
Swedz: Also used for my tesla tower :P I need to unload the network when the machine is destroyed in world

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
Technici4n: So how are you doing that? Replacing the `shapeMatcher` in `onLink()` makes reasoning about the code too hard, so I hope that's not what you're doing.

src/main/java/aztech/modern_industrialization/machines/multiblocks/ShapeMatcher.java
Swedz: I use them in my [custom shape matcher](https://github.com/Swedz/Extended-Industrialization/blob/tesla-tower/src/main/java/net/swedz/extended_industrialization/machines/blockentity/multiblock/teslatower/SameCableTierShapeMatcher.java) to prevent the tesla tower from linking when there is more than one tier of energy hatch present

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
Technici4n: Ok that's reasonable. :P

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
Technici4n: This might be better achieved by adding the ability to perform an additional check in the `ShapeMatcher`.

Swedz: No, I use the createShapeMatcher method I added :p

src/main/java/aztech/modern_industrialization/machines/multiblocks/ShapeMatcher.java
Technici4n: Ah I see. It's probably better to add an optional additional condition to the shape matcher - possibly in a protected method that returns true by default - than overriding half of the class.

Swedz: The main reason I override the rematch method is because it prevents needlessly having to iterate over the same stuff twice but I suppose it's such a small dataset it doesn't matter too much. I can make that change

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
Technici4n: So you don't use `onLink`? Let's get rid of it then.

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
Technici4n: I prefer a reset in `onRematch()` than separate handling of `onMatchFailure()`.

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
Swedz: I do use onLink and onUnlink. I use it to register and unregister additional listeners sort of in the same way MI does it (just immediately after). Would you prefer I get rid of these two methods and instead made a custom ShapeMatcher that adds to the un/registerListeners methods?

Technici4n: Yes let's do that

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
Swedz: ~~This needs to be accessible in subclasses to check for success/failure or to get the matched hatches in onRematch~~

src/main/java/aztech/modern_industrialization/machines/multiblocks/MultiblockMachineBlockEntity.java
Swedz: I use it to check if my ShapeMatcher (SameCableTierShapeMatcher) has mismatching hatches or not and display info about it in the multiblock GUI (see [here](https://github.com/Swedz/Extended-Industrialization/blob/6b730612783d5d0dd8376044dc9bd1e88056bd16/src/main/java/net/swedz/extended_industrialization/machines/blockentity/multiblock/teslatower/TeslaTowerBlockEntity.java#L128-L131)). Mismatching hatches in this case refers to if there are energy hatches that aren't the same cable tier as others.

Technici4n: I'd probably change that info to be stored as a field in the BE, updated in a callback. I don't want the shape matcher to be too accessible.

src/main/java/aztech/modern_industrialization/machines/multiblocks/ShapeMatcher.java
Swedz: Maybe the better option is moving the simpleMembers loop to a checkRematch method? That way the members don't need to be iterated over more than once and still avoids needing to override rematch.

Technici4n: What makes you think that iterating over the members a second time is a potential performance issue?

Swedz: Depends on the number of members but I know some people have made some rather large multiblocks. Requesting world info about block states and block entities more than once for the same check is kind of unnecessary, no?

Technici4n: Depends what you do with it. If you're just checking for hatch tiers, it should be enough to iterate over `matchedHatches`.

Technici4n (approved):

PR #939, opened 2024-11-25
Make CableTier more accessible

Swedz:
Contains some parts of #900 

It might make sense to also add the `CableTierHolder` interface to electric machines and get the tier from their casing component. That can easily be done. Doing it for multi blocks is a little ambiguous since a multi block electric machine can have multiple types of hatches on it. It could just use the highest tier, though. Would either of these changes be acceptable?

### DIFF
--- a/src/main/java/aztech/modern_industrialization/machines/blockentities/hatches/EnergyHatch.java
+++ b/src/main/java/aztech/modern_industrialization/machines/blockentities/hatches/EnergyHatch.java
@@ -25,6 +25,7 @@ package aztech.modern_industrialization.machines.blockentities.hatches;
 
 import aztech.modern_industrialization.MICapabilities;
 import aztech.modern_industrialization.api.energy.CableTier;
+import aztech.modern_industrialization.api.energy.CableTierHolder;
 import aztech.modern_industrialization.api.energy.EnergyApi;
 import aztech.modern_industrialization.api.energy.MIEnergyStorage;
 import aztech.modern_industrialization.api.machine.holder.EnergyComponentHolder;
@@ -39,12 +40,13 @@ import aztech.modern_industrialization.machines.multiblocks.HatchType;
 import java.util.List;
 import net.minecraft.world.level.block.entity.BlockEntityType;
 
-public class EnergyHatch extends HatchBlockEntity implements EnergyComponentHolder {
+public class EnergyHatch extends HatchBlockEntity implements EnergyComponentHolder, CableTierHolder {
 
     public EnergyHatch(BEP bep, String name, boolean input, CableTier tier) {
         super(bep, new MachineGuiParameters.Builder(name, false).build(), new OrientationComponent.Params(!input, false, false));
 
         this.input = input;
+        this.tier = tier;
 
         this.energy = new EnergyComponent(this, 30 * 20 * tier.getEu());
         insertable = energy.buildInsertable((CableTier tier2) -> tier2 == tier);
@@ -55,7 +57,8 @@ public class EnergyHatch extends HatchBlockEntity implements EnergyComponentHold
         this.registerComponents(energy);
     }
 
-    private final boolean input;
+    protected final boolean input;
+    protected final CableTier tier;
 
     protected final EnergyComponent energy;
     protected final MIEnergyStorage insertable;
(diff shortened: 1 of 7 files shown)


Technici4n:
What is the motivation behind these changes? I.e. how are they helpful?

Swedz:
I use it quite a bit for my tesla coil machines I've been working on for EI.

It lets me make an anonymous instance of CasingComponent to track when the casing changes and do some updates to the network when that happens.

Plus, it makes grabbing the cable tier of a tesla coil block/multiblock much more straightforward (I was previously using a mixin to do this) - and I need that to be able to wirelessly transfer power on a specific voltage.

Technici4n:
> It lets me make an anonymous instance of CasingComponent to track when the casing changes and do some updates to the network when that happens.

It would be cleaner to add an optional change notification callback IMO.

src/main/java/aztech/modern_industrialization/machines/components/CasingComponent.java
```diff
@@ -42,15 +43,15 @@
 import net.minecraft.world.level.block.Blocks;
 import org.jetbrains.annotations.Nullable;
 
-public class CasingComponent implements IComponent, DropableComponent {
+public class CasingComponent implements IComponent, DropableComponent, CableTierHolder {
```
Technici4n: This is not how the holder is meant to be used IMO. A check for `blockEntity instanceof CableTierHolder` will never succeed like this.

src/main/java/aztech/modern_industrialization/machines/components/CasingComponent.java
```diff
@@ -42,15 +43,15 @@
...
-    private ItemStack casingStack = ItemStack.EMPTY;
-    private CableTier currentTier = CableTier.LV;
+    protected ItemStack casingStack = ItemStack.EMPTY;
+    protected CableTier currentTier = CableTier.LV;
 
     /**
      * Sets the current casing stack and update {@link #currentTier} accordingly.
      */
-    private void setCasingStack(ItemStack stack) {
+    protected void setCasingStack(ItemStack stack) {
```
Technici4n: Can you replace these changes by an optional change notification callback?

Technici4n (changes requested):

src/main/java/aztech/modern_industrialization/machines/components/CasingComponent.java
Swedz: What do you think about adding the holder interface to single block electric machines? And then what about multiblocks? How it would work for multiblocks is a bit ambiguous. I think keeping the getCasingTier method for the CasingComponent is good even if the interface is removed from it.

Technici4n: Yes we can add it to electric machines. Multiblocks can probably wait until someone requests them. Yes to keep the method without the interface.

Swedz:
I made those changes. I didn't add CableTierHolder to `AbstractStorageMachineBlockEntity` (and thus `TransformerMachineBlockEntity`) because it holds two different cable tiers and I wasn't sure which one would be best to return there.

src/main/java/aztech/modern_industrialization/machines/blockentities/hatches/EnergyHatch.java
```diff
@@ -55,7 +57,8 @@ public EnergyHatch(BEP bep, String name, boolean input, CableTier tier) {
         this.registerComponents(energy);
     }
 
-    private final boolean input;
+    protected final boolean input;
```
Technici4n: Any reason not to keep them `private`?

Swedz: Whoops, missed that one

Swedz: Whats the reason for these to be private when the other fields are protected, out of curiosity?

Technici4n: I think you're asking the wrong question, and that the right question would be: Why are the others `protected` even though they don't seem to be accessed from outside the class? 😄 No idea though

Swedz: Fair enough. Trying to think of a situation where I may want to extend EnergyHatch and use those fields, I don't think there's really any that don't have alternatives. I can swap these to private and come back in the future if I wind up needing it for some reason.

Technici4n (approved):

PR #940, opened 2024-11-25
Allow vertical orientations with component

Swedz:
Contains some parts of #900

### DIFF
--- a/src/main/java/aztech/modern_industrialization/machines/components/OrientationComponent.java
+++ b/src/main/java/aztech/modern_industrialization/machines/components/OrientationComponent.java
@@ -89,7 +89,7 @@ public class OrientationComponent implements IComponent {
                 return true;
             }
         } else {
-            if (face.getAxis().isHorizontal()) {
+            if (params.canBeVertical || face.getAxis().isHorizontal()) {
                 facingDirection = face;
             }
             // We consume the event to prevent the GUI from opening.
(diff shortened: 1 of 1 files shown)


Technici4n:
I don't think the models in MI would be too happy with this?

Swedz:
I'd need to double check later, but this change only allows for machines to explicitly allow vertical orientations and by default wouldn't be any different. My goal for this is to update the Machine Chainer from EI to let it face up and down. I can give more info on this when I get back to my desk later tonight

Technici4n:
Would setting `canBeVertical` to `true` not cause problems with MI's models? If yes just add a comment that it needs a custom model. I'd also add a comment that this is for addons, but that MI itself does not use it. (So it doesn't get removed accidentally).

Swedz:
Yes it would, I made a custom loader for the machine chainer to be able to use this new flag. Its part of why I wanted to revamp a bit of the machine model stuff in #900 as well. It lets me extend MachineBakedModel and MachineOverlaysJson to easily get a lot of the same functionality while also not having to copy paste a large part of the MI code. I can add a comment making this clear. If you're at all curious, [here](https://github.com/Swedz/Extended-Industrialization/blob/c5d24c937dde8f2f100c6340eda4a7a88021a862/src/main/java/net/swedz/extended_industrialization/client/model/chainer/MachineChainerBakedModel.java) is what my machine chainer's baked model looks like using these changes, if anything to get an idea of how this is used.

Technici4n (approved):

PR #941, opened 2024-11-25
Let BackgroundRenderedSlots override the background atlas

Swedz:
Contains some parts of #900

### DIFF
--- a/src/client/java/aztech/modern_industrialization/machines/gui/MachineScreen.java
+++ b/src/client/java/aztech/modern_industrialization/machines/gui/MachineScreen.java
@@ -211,10 +211,15 @@ public class MachineScreen extends MIHandledScreen<MachineMenuClient> implements
 
     private void renderConfigurableSlotBackgrounds(GuiGraphics guiGraphics) {
         for (Slot slot : this.menu.slots) {
-            if (slot instanceof BackgroundRenderedSlot brs) {
+            if (slot.isActive() && slot instanceof BackgroundRenderedSlot brs) {
                 int px = leftPos + slot.x - 1;
                 int py = topPos + slot.y - 1;
-                guiGraphics.blit(SLOT_ATLAS, px, py, brs.getBackgroundU(), brs.getBackgroundV(), 18, 18);
+                if (slot.getItem().isEmpty()) {
+                    var atlas = brs.getBackgroundAtlasLocation();
+                    guiGraphics.blit(atlas == null ? SLOT_ATLAS : atlas, px, py, brs.getBackgroundU(), brs.getBackgroundV(), 18, 18);
+                } else {
+                    guiGraphics.blit(SLOT_ATLAS, px, py, 0, 0, 18, 18);
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/client/java/aztech/modern_industrialization/machines/gui/MachineScreen.java
```diff
@@ -211,10 +211,15 @@ protected void renderBg(GuiGraphics guiGraphics, float delta, int mouseX, int mo
 
     private void renderConfigurableSlotBackgrounds(GuiGraphics guiGraphics) {
         for (Slot slot : this.menu.slots) {
-            if (slot instanceof BackgroundRenderedSlot brs) {
+            if (slot.isActive() && slot instanceof BackgroundRenderedSlot brs) {
                 int px = leftPos + slot.x - 1;
                 int py = topPos + slot.y - 1;
-                guiGraphics.blit(SLOT_ATLAS, px, py, brs.getBackgroundU(), brs.getBackgroundV(), 18, 18);
+                if (slot.getItem().isEmpty()) {
```
Technici4n: Why this added check? The default values can be accessed through `BackgroundRenderedSlot.super.<method name>()`.

src/main/java/aztech/modern_industrialization/inventory/BackgroundRenderedSlot.java
```diff
@@ -23,10 +23,21 @@
...
+import org.jetbrains.annotations.Nullable;
+
 /**
  * Implement this on a slot to render its background automatically.
  */
 public interface BackgroundRenderedSlot {
+    /**
+     * @return the {@link ResourceLocation} of the slot atlas texture to use for the slot background. When null, uses the MI slot atlas
+     */
+    @Nullable
```
Technici4n, line 37: Instead of this being nullable, it could return the MI slot atlas.

Swedz: That is true

src/client/java/aztech/modern_industrialization/machines/gui/MachineScreen.java
Swedz: The added isEmpty check is so that the standard slot texture is rendered when an item is in the slot, but if there isn't an item, the designated slot texture is rendered instead. This makes it so that if the slot texture doesn't fully align with all items it may accept, it wont still render it behind the item if there is an item there.

As for the isActive check, I can't find my use case for it... it is a logical check to have there regardless I think. I may have had a need for it at one point and then changed something, not sure.

Swedz: I don't think the default values can be accessed using `super` here?

Technici4n: Checking for isActive is a bad idea I think, maybe some background is still needed? Could make it so that returning a null atlas skips rendering if you want.

Technici4n: And same for isEmpty, I'd rather give full control to the slot. For example fluid slots are always empty I think.

Swedz: A null atlas is probably a good way to do it. You make a good point about isEmpty always being false for fluid slots. Leaving it entirely up to the slot is fine, since you can still control whether it shows or not through that.

PR #943, opened 2024-11-26
Fix #933: Add item tags to wires

### LINKED ISSUE #933: Wire Tags
Please add the `c:wires` and `c:wires/{material}` item tags to wires for unification purposes.

Swedz:
I had to make `CATEGORY_TAGS` get defined in a static block because `Map.of(...)` only goes up to 10 entries but otherwise this was simple enough


Technici4n:
Did you check that other mods have a consistent ingot to wire ratio? And did you check the non-generated recipes? I'm surprised that none of them changed.

Swedz:
Honestly, I don't even know what mod the person has that uses c:wires and they want to unify with. I was looking for a while and didn't find any. I'll ask on their issue to be sure

PR #945, opened 2024-11-27
Make machine casings use ResourceLocations and translations

Swedz:


src/client/java/aztech/modern_industrialization/machines/models/MachineBakedModel.java
```diff
@@ -57,22 +57,25 @@ public class MachineBakedModel implements IDynamicBakedModel {
     public static final String CASING_FOLDER = "machine_casing";
 
     public static ModelResourceLocation getCasingModelId(MachineCasing casing) {
-        return ModelResourceLocation.standalone(MI.id(CASING_FOLDER + "/" + casing.name));
+        return ModelResourceLocation
```
Technici4n: Looks like a good use case for `casing.key.withPath(p -> CASING_FOLDER + "/" + p)`.

docs/ADDING_MACHINES.md
```diff
@@ -327,12 +327,12 @@ remember that the top, side and bottom textures of a casing must be `modern_indu
 For example:
 ```js
 MIMachineEvents.registerCasings(event => {
-    // Register two casings.
+    // Register a casing.
     // This doesn't register any model! Either add models or add the top/side/bottom textures.
-    event.register("my_fancy_casing", "my_other_casing");
+    event.register("fancy_casing", "Fancy");
```
Technici4n: This is a breaking change for KubeJS users, might be best to avoid it?

docs/ADDING_MACHINES.md
```diff
@@ -327,12 +327,12 @@ remember that the top, side and bottom textures of a casing must be `modern_indu
...
 
     // This registers a new casing with the same model as a diamond block!
-    event.registerBlockImitation("my_diamond_casing", "minecraft:diamond_block");
+    event.registerBlockImitation("diamond_casing", "Diamond", "minecraft:diamond_block");
```
Technici4n: We can use the display name of the block in this case, no?

src/client/java/aztech/modern_industrialization/machines/models/MachineModelBaker.java
```diff
@@ -0,0 +1,34 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package aztech.modern_industrialization.machines.models;
+
+import java.util.Map;
+import net.minecraft.client.renderer.texture.TextureAtlasSprite;
+import net.minecraft.resources.ResourceLocation;
+
+public interface MachineModelBaker {
```
Technici4n: All of this stuff looks unnecessary.

Technici4n (changes requested):

docs/ADDING_MACHINES.md
Kazuhiko-Gushiken: I agree with this, it threw off my casing code first time around.

docs/ADDING_MACHINES.md
Swedz: Not sure what would be a good solution for this, the translation name for casings are used in the tooltips of blocks in the structure PR I've been working on.

docs/ADDING_MACHINES.md
Swedz: I'm not seeing a clear solution for this. The english name of the block would need to be pulled from somewhere for datagen. Not sure if that is guaranteed available, especially on a server.

Technici4n (approved):

PR #985, opened 2024-12-27
Use new NeoForge attribute API for quantum armor, quantum sword and gravichestplate

### LINKED ISSUE #984: Two "when on body" sections in the tooltip for Quantum Chestplate
When Modern Industrialization 2.2.30 is installed alongside Apothic Attributes 2.5.0 in NeoForge 21.1.90 and Minecraft 1.21.1, the Quantum Chestplate has two sections in its tooltip labeled "When on Body":
image

The first section is added by the mod itself and contains the specially-rendered armor attribute describing the Quantum Armor's bonus, while the second section is added by vanilla to show the other attribute modifiers added by Apothic Attributes and other mods.

Swedz:
Closes #984

Drops the `ItemStackMixin` mixin because `addAttributeTooltips` is now deprecated and no longer called.

Requires NeoForge version to get bumped to 21.1.60 because that is when the attribute API was added, which simultaneously caused the issue of #984 to appear.


src/main/java/aztech/modern_industrialization/attributes/QuantumArmorAttribute.java
```diff
@@ -0,0 +1,46 @@
...
+import net.minecraft.world.item.TooltipFlag;
+
+public class QuantumArmorAttribute extends DisplayNamedAttribute {
+    public QuantumArmorAttribute() {
+        super(MIText.AttributeQuantumArmor.getTranslationKey(), 0);
+    }
+
+    @Override
+    public MutableComponent toValueComponent(AttributeModifier.Operation operation, double value, TooltipFlag flag) {
+        return Component.literal("\u00B9\u2044\u2084 |\u221E> + \u00B3\u2044\u2084 |0>");
```
James103, line 39: The hardcoded string only makes sense when dealing with items which grant the bonus of one piece of quantum armor. However, attribute viewers such as the one added by Apothic Attributes may render the quantum armor attribute with a value that is not exactly 1; the other common value being 4 when wearing a full set of Quantum Armor. The same issue can occur with custom items.

Swedz: Neoforge handles this for us, so not really an issue
image

Technici4n:
Cool, thank you!

PR #1021, opened 2025-02-07
Stop item pipes from self-inserting their own extractions

Swedz:
This just adds an identity check to prevent pipe nodes from inserting into themselves.


Swedz:
Discussing in the dev channel how to improve the UI for this. Will mark as ready to merge when I feel like its good

Technici4n:
I was gonna complain about the option, then I realized this PR simply disables self-insertion. That's great! :smile:

Can you update the title and description, and make sure that there is no mention of self-insertion in the guide book? (Not sure there was tbh, I don't know)

Swedz:
Yeah we had discussed having it being an option on the discord but concluded it wasn't necessary. I'll update the title now. I can do some searches in the guide (not at this moment) to see if I can find any mention of this and comment back here.

Swedz:
I'm not seeing any mention of item pipes self inserting in the book from my searches. Should be all set to merge

PR #1022, opened 2025-02-09
Pipe Config Card to extract blocks from items containing items (such as Banks from Bank Storage)

Swedz:
A few things to comment on:

- The camouflage applying is now done entirely server-side, since item handler capabilities shouldn't be messed with on the client side. In the case of Bank Storage, they use a read-only item handler capability on the client side. Because of this, the "No _____ in inventory" actionbar was being displayed and sounds wouldn't always play because the client got a different result when extracting than the server did.

- `TransferHelper#extractMatching` had to get adjusted to allow for recursive item extracting.

  - It won't go any deeper than one level though, so no backpacks inside backpacks - for example. This might be worth changing? Would be as simple as passing `true` to the recursive calls.
  
  - `ItemStack#matches` checks against item count which is (likely?) not intended here, changed it to `ItemStack#isSameItemSameComponents` to retain functionality while also fixing this.

### DIFF
--- a/src/main/java/aztech/modern_industrialization/util/TransferHelper.java
+++ b/src/main/java/aztech/modern_industrialization/util/TransferHelper.java
@@ -30,6 +30,7 @@ import net.minecraft.world.item.ItemStack;
 import net.neoforged.neoforge.capabilities.Capabilities;
 import net.neoforged.neoforge.items.IItemHandler;
 import net.neoforged.neoforge.items.ItemHandlerHelper;
+import net.neoforged.neoforge.items.wrapper.PlayerInvWrapper;
 
 public class TransferHelper {
     public static void moveAll(IItemHandler src, IItemHandler target, boolean stackInTarget) {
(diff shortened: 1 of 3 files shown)


src/main/java/aztech/modern_industrialization/util/TransferHelper.java
```diff
@@ -77,19 +78,32 @@ public static ItemStack extractMatching(IItemHandler src, Predicate<ItemStack> p
             var stack = src.getStackInSlot(slot);
             if (predicate.test(stack)) {
                 ret = src.extractItem(slot, maxAmount, false);
+            } else if (recursive) {
+                var capability = stack.getCapability(Capabilities.ItemHandler.ITEM);
```
Technici4n: I don't think that's correct: the stack coming from `getStackInSlot` should not be modified in any way. Might be fine for player inventories, but then it shouldn't be in a general purpose method like this.

I think that trying to extract from the item inside the item handler is too complex, so maybe this should be rewritten somehow to operate at the `Inventory` level?

src/main/java/aztech/modern_industrialization/util/TransferHelper.java
```diff
@@ -77,19 +78,32 @@ public static ItemStack extractMatching(IItemHandler src, Predicate<ItemStack> p
...
             return ItemStack.EMPTY;
         }
+        final ItemStack finalRet = ret;
 
         // Try to extract more
         ++slot;
         for (; slot < srcSlots && maxAmount < ret.getCount(); ++slot) {
             var stack = src.getStackInSlot(slot);
-            if (ItemStack.matches(stack, ret)) {
+            if (ItemStack.isSameItemSameComponents(stack, ret)) {
```
Technici4n, line 123: Good catch 😅

Swedz:
Should be all good now. Noticed a couple other issues with `extractMatching`'s second iteration that this PR also fixes now:
- Had `simulate` set to `true` when extracting from an `IItemHandler` so it was never *really* pulling extra items after the first stack
- The if-statement portion of the for loop had `maxAmount < ret.getCount()` which was the wrong way around. Only should search for additional items if the amount grabbed so far (`ret`) is less than the desired amount
- Before the iteration, `slot` was being incremented again, however that is already done at the end of the first iteration. this extra increment on that variable was making it skip the slot immediately after the slot the first stack is pulled from

src/main/java/aztech/modern_industrialization/util/TransferHelper.java
```diff
@@ -67,32 +69,69 @@ public static void moveAll(IItemHandler src, IItemHandler target, boolean stackI
...
 
-    public static ItemStack extractMatching(IItemHandler src, Predicate<ItemStack> predicate, int maxAmount) {
-        int srcSlots = src.getSlots();
+    private interface ItemExtractor {
+        ItemStack get(int slot);
+
+        ItemStack extract(int slot, int amount);
+    }
 
+    private static ItemStack extractMatching(int srcSlots, ItemExtractor extractor, Predicate<ItemStack> predicate, int maxAmount,
```
Technici4n: I don't like this, it's too complicated. Can you not do a first pass through the player inventory that will try to extract from the item-containing item and then a second pass calling the unmodified (except for bug fixes) `extractMatching`? As a side note, be careful with item-containing items of size > 1, it's a big mess and the handling is really ugly.

Technici4n (changes requested):

src/main/java/aztech/modern_industrialization/util/TransferHelper.java
Swedz: Yes I can do that. It just means copying the old `extractMatching` and having duplicate code though, which isn't ideal. I'll make sure it excludes `IItemHandler` items with > 1 item, good point.

Swedz: Never mind that, I forgot `PlayerInvWrapper` exists.

Technici4n: Hmm indeed. Duplication is not great either.

Technici4n: But what is there to duplicate? What I have in mind is just to loop through every slot in `Inventory` and check if it contains a container item. If yes get the capability and call `extractMatching` on it. (Not sure we need to set the stack again; I don't think so).

Swedz: I've got what I think is a nifty solution, I'll push it in a few mins after I test it

src/main/java/aztech/modern_industrialization/util/TransferHelper.java
```diff
@@ -67,6 +70,37 @@ public static void moveAll(IItemHandler src, IItemHandler target, boolean stackI
...
+                predicate = other -> ItemStack.isSameItemSameComponents(finalRet, other);
+            }
+            for (int slot = 0; slot < srcSlots && maxAmount > ret.getCount(); ++slot) {
+                var stack = inventory.getItem(slot);
+                if (stack.getCount() != 1) {
+                    continue;
+                }
+                var capability = stack.getCapability(Capabilities.ItemHandler.ITEM);
+                if (capability != null) {
+                    var extracted = extractMatching(capability, predicate, maxAmount);
```
Technici4n:
```suggestion
                    var extracted = extractMatching(capability, predicate, maxAmount - ret.getCount());
```

Technici4n (approved):

PR #1023, opened 2025-02-09
Make Multiblock Materials recipe category have scrollable/paged views to allow for more blocks to be shown

Swedz:
This only properly supports EMI and JEI. REI just gets an extra row of slots for now since it was being difficult :P

The number of rows/columns can be adjusted if that's desirable. I thought 3 rows looked nice, but it took up quite a bit of space on the screen. It might not really matter since its not often you're looking at multiple multiblocks shapes at a time anyway. With 2 rows (how it is now), it lets you see 3 shapes per page in the category.


pietro-lopes:
Following this

Technici4n (approved):
Looks good after a bit of testing. We'll do the rest of the testing in production. @pietro-lopes let me know if there's any issue I guess.

PR #1025, opened 2025-02-17
Add FE wires

Swedz:


src/generated/resources/assets/modern_industrialization/lang/en_us.json
```diff
@@ -1456,13 +1473,17 @@
...
+  "tag.modern_industrialization.fe_cables_part": "FE Cables Part",
+  "tag.modern_industrialization.fe_wires": "FE Wires",
   "tag.modern_industrialization.fluid_pipes": "Fluid Pipes",
   "tag.modern_industrialization.forge_hammer_tools": "Forge Hammer Tools",
   "tag.modern_industrialization.item_pipes": "Item Pipes",
   "tag.modern_industrialization.me_wires": "ME Wires",
   "tag.modern_industrialization.replicator_blacklist": "Replicator Blacklist",
   "tag.modern_industrialization.tanks": "Tanks",
   "text.autoconfig.modern_industrialization.option.armorHudYPosition": "Space between the top of the screen and the Jetpack/GraviChestPlate overlay text.",
+  "text.autoconfig.modern_industrialization.option.baseFEWireTransfer": "Base amount of FE transferred by fe wires every tick.",
```
James103:
```suggestion
  "text.autoconfig.modern_industrialization.option.baseFEWireTransfer": "Base amount of FE transferred by FE wires every tick.",
```

src/main/java/aztech/modern_industrialization/MIConfig.java
```diff
@@ -88,6 +91,8 @@ public class MIConfig implements ConfigData {
     public int armorHudYPosition = 4;
     @EnglishTranslation(value = "Base amount of items transferred by item pipes every 3 seconds.")
     public int baseItemPipeTransfer = 16;
+    @EnglishTranslation(value = "Base amount of FE transferred by fe wires every tick.")
```
James103:
```suggestion
    @EnglishTranslation(value = "Base amount of FE transferred by FE wires every tick.")
```

Swedz:
I do think this would be a nice feature. If you are not interested in it, feel free to close this and I will look into potentially including it in an addon. Otherwise let me know and I will resolve the merge conflicts.

Kazuhiko-Gushiken:
Is this just FE counterparts of EU cables? Same, 3 in a block, different power transfer tiers?

Swedz:
Yes. Transfer rates work similar to item pipes where you need upgrades to increase throughput

Kazuhiko-Gushiken:
Ohh that's pretty cool. I think that'd be a good addition, though maybe a config to disable it (have it enabled by default) would be nice? Just in case. Also, maybe a tooltip? If bidirectional compat is off, the tooltip states "Cannot be used to power MI machines." but if bidirectional compat is on, the tooltip doesnt exist (though i guess it would be nice to have it auto disable if bidirectional compat is on, but people may like to use it over the regular cables, but then there's two cables where one is pointless to use when the other exists.)

Swedz:
A config option is already included in the PR. A tooltip mentioning it can't be used for MI machines when bidirectional compat is off would be nice, yeah.

PR #1055, opened 2025-03-27
Add support for fluid tags and ingredients in MIRecipeJson

Swedz:


src/main/java/aztech/modern_industrialization/machines/recipe/MIRecipeJson.java
```diff
@@ -125,12 +126,21 @@ public T addItemOutput(ItemVariant variant, int amount, float probability) {
...
 
-    public T addFluidInput(String fluid, int amount) {
-        return addFluidInput(fluid, amount, 1);
+    public T addFluidInput(String maybeTag, int amount) {
+        return addFluidInput(maybeTag, amount, 1);
     }
 
-    public T addFluidInput(String fluid, int amount, float probability) {
-        return addFluidInput(BuiltInRegistries.FLUID.get(ResourceLocation.parse(fluid)), amount, probability);
+    public T addFluidInput(String maybeTag, int amount, float probability) {
```
Technici4n: Do you plan to use this `maybeTag` function? If not don't add it. It's an old pattern, nowadays I would have used an `Ingredient` from the start.

Swedz: Fair enough, I thought it would be used by KJS but after checking it seems like that's not necessary. I'll remove this

PR #1128, opened 2025-06-19
Add Almost Unified integration

rlnt:
### Information

This pull request adds Almost Unified integration. The dependency is added as a soft compile-time dependency.
The integration has been gated behind a startup config flag.

### Reasoning

The current implementation of slot locking in machines only works with the recipe ID since the recipe transfer handlers from recipe viewers only send the ID to the server. Because a slot can only be locked to a single item, the algorithm to find a suitable slot item needs to be improved when Almost Unified is present. In the current release version, MI defaults to its own items if the player has no suitable items in their inventory. This can result in locked items that are unobtainable due to unification.

Issue reference: https://github.com/AllTheMods/ATM-10/issues/1667

### Changes

- added the Almost Unified API as a compile-time dependency
  - we push to BlameJared maven, but since ModMaven mirrors that, I didn't want to add another maven repository
  - I noticed there is a duplicate maven entry, which I removed
- added abstraction for guarded access to the Almost Unified API with the facade approach other integrations use
- added a startup config flag to disable the Almost Unified integration
- changed locking logic to fetch the preferred item from Almost Unified if present
  - I left the inventory logic as the first approach because I think that makes more sense, feel free to change that
  - the API returns null if there is no replacement item, and MI will use the old behavior in that case
- added an Almost Unified plugin and unifier for MI recipes

### Plugin and Unifier

The new plugin allows defining custom recipe unifiers. Unifiers tell Almost Unified how specific recipes need to be treated if they are not covered by our generic recipe unifier. We currently host the recipe unifier for MI in Almost Unified itself ([reference](https://github.com/AlmostReliable/almostunified/blob/1.21.1/NeoForge/src/main/java/com/almostreliable/unified/compat/unification/ModernIndustrializationRecipeUnifier.java). We would love to move it to first-party, so you can adjust the unifier if the recipe structure changes in the future. Our plugin system is built so that modded recipe unifiers have priority over the built-in ones. And because MI uses the same recipe structure for all recipes, the unifier is very simple.

If you don't want to have the unifier yourself, please let me know. I thought it would be a good point in time to move it, since I added integration anyway.

CC: @Swedz If this merge request is merged and the unifier is moved to MI, you would need to update your add-on and register the MI unifier for your add-on as well.


gradle.properties
```diff
@@ -49,3 +49,4 @@ jei_version_range=[19.19.0,]
 kubejs_version=2101.7.1-build.181
 rei_version=16.0.729
 rei_annotations_version=12.0.93
+au_version=1.2.5
```
Technici4n: Should be sorted alphabetically but I can do it myself.

Technici4n:
Yeah the unifier is fine. What about forge hammer recipes?

rlnt:
Looking at the codec of the Forge Hammer recipe, it only has an ingredient under the `ingredient` key and an item stack under the `result` key. Both of these keys are covered by our generic recipe unifier, which is still applied in the custom unifier.
https://github.com/AztechMC/Modern-Industrialization/blob/1.21.x/src/main/java/aztech/modern_industrialization/blocks/forgehammer/ForgeHammerRecipe.java

Technici4n (approved):
Thanks!

PR #1140, opened 2025-07-14
Allow custom multiblocks to have a custom CrafterComponent

Luligabi1:
# Problem

`AbstractCraftingMultiblockBlockEntity` implements `CrafterComponentHolder`, but directly accesses `crafter` instead of calling `getCrafterComponent` on `tick`, meaning any implementing classes are unable to have a custom `CrafterComponent`. 

# Solution

All direct accesses to `crafter` on `tick` are replaced by `getCrafterComponent`. This PR also includes a new method called `onCraft` that allows custom `CrafterComponent`s to detect when a recipe is completed without overriding/mixin injecting `tickRecipe`.

# Use case

Imagine a machine that alters the non-consumed input stack in some way (i.e. one of its data components' value). If said way is not recipe dependent, `CrafterComponent` is the most sensible place to modify it.

### DIFF
--- a/src/main/java/aztech/modern_industrialization/machines/blockentities/multiblocks/AbstractCraftingMultiblockBlockEntity.java
+++ b/src/main/java/aztech/modern_industrialization/machines/blockentities/multiblocks/AbstractCraftingMultiblockBlockEntity.java
@@ -94,17 +94,17 @@ public abstract class AbstractCraftingMultiblockBlockEntity extends MultiblockMa
             boolean newActive = false;
 
             if (operatingState == OperatingState.TRYING_TO_RESUME) {
-                if (crafter.tryContinueRecipe()) {
+                if (getCrafterComponent().tryContinueRecipe()) {
                     operatingState = OperatingState.NORMAL_OPERATION;
                 }
             }
 
             if (operatingState == OperatingState.NORMAL_OPERATION) {
-                if (crafter.tickRecipe()) {
+                if (getCrafterComponent().tickRecipe()) {
                     newActive = true;
                 }
             } else {
-                crafter.decreaseEfficiencyTicks();
+                getCrafterComponent().decreaseEfficiencyTicks();
             }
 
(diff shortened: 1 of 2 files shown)


Technici4n:
Oh so you're writing an addon? Is it public yet?

src/main/java/aztech/modern_industrialization/machines/components/CrafterComponent.java
```diff
@@ -184,6 +184,9 @@ public void increaseEfficiencyTicks(int increment) {
         efficiencyTicks = Math.min(efficiencyTicks + increment, maxEfficiencyTicks);
     }
 
+    protected void onCraft() {
```
Technici4n: Do you need a subclass of `CrafterComponent`, or would an `onCraft` callback in the Behavior be more appropriate?

Luligabi1: I'll probably do a subclass, but that'd be a better place either way, yeah

Luligabi1:
> Oh so you're writing an addon? Is it public yet?

Hopefully I can get it out this month 🙂

src/main/java/aztech/modern_industrialization/machines/blockentities/multiblocks/AbstractCraftingMultiblockBlockEntity.java
```diff
@@ -94,17 +94,17 @@ public final void tick() {
             boolean newActive = false;
 
             if (operatingState == OperatingState.TRYING_TO_RESUME) {
-                if (crafter.tryContinueRecipe()) {
+                if (getCrafterComponent().tryContinueRecipe()) {
```
Technici4n: Do you need to make a CrafterComponent subclass still?

If yes -> make the `crafter` field private
If no -> we don't even need changes to this file anymore

Technici4n (approved):

PR #1159, opened 2025-08-16
Allow for custom hatch types

Swedz:
This changes `HatchType` from being an enum to being a class structured similarly to `MachineCasing` so addons can add custom hatch types.

### DIFF
--- a/src/main/java/aztech/modern_industrialization/machines/multiblocks/HatchFlags.java
+++ b/src/main/java/aztech/modern_industrialization/machines/multiblocks/HatchFlags.java
@@ -40,6 +40,10 @@ public class HatchFlags {
         return allowed.contains(type);
     }
 
+    public Set<HatchType> values() {
+        return allowed;
+    }
+
     public static class Builder {
         private final Set<HatchType> allowed = new HashSet<>();
 
--- a/src/main/java/aztech/modern_industrialization/machines/multiblocks/HatchFlags.java
+++ b/src/main/java/aztech/modern_industrialization/machines/multiblocks/HatchFlags.java
@@ -23,24 +23,28 @@
  */
 package aztech.modern_industrialization.machines.multiblocks;
 
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
 public class HatchFlags {
     public static final HatchFlags NO_HATCH = new Builder().build();
 
-    private final int flags;
+    private final Set<HatchType> allowed;
 
-    public HatchFlags(int flags) {
-        this.flags = flags;
+    public HatchFlags(Set<HatchType> allowed) {
+        this.allowed = Collections.unmodifiableSet(allowed);
     }
 
     public boolean allows(HatchType type) {
-        return (flags & (1 << type.getId())) > 0;
+        return allowed.contains(type);
     }
 
     public static class Builder {
-        private int flags = 0;
+        private final Set<HatchType> allowed = new HashSet<>();
 
         public Builder with(HatchType type) {
-            flags |= 1 << type.getId();
+            allowed.add(type);
             return this;
         }
 
@@ -53,7 +57,7 @@ public class HatchFlags {
         }
 
         public HatchFlags build() {
-            return new HatchFlags(flags);
+            return new HatchFlags(allowed);
         }
     }
 }
--- a/src/main/java/aztech/modern_industrialization/machines/multiblocks/HatchFlags.java
+++ b/src/main/java/aztech/modern_industrialization/machines/multiblocks/HatchFlags.java
@@ -23,7 +23,6 @@
  */
 package aztech.modern_industrialization.machines.multiblocks;
 
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
@@ -33,7 +32,7 @@ public class HatchFlags {
     private final Set<HatchType> allowed;
 
     public HatchFlags(Set<HatchType> allowed) {
-        this.allowed = Collections.unmodifiableSet(allowed);
+        this.allowed = Set.copyOf(allowed);
     }
 
     public boolean allows(HatchType type) {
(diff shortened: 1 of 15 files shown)


Swedz:
Note that this does introduce breaking changes for addons by moving the `HatchType` instances to a `HatchTypes` class. I'm not too upset by that, but others might be... just a heads up. This can be changed if deemed important.

src/main/java/aztech/modern_industrialization/machines/multiblocks/HatchTypes.java
```diff
@@ -0,0 +1,85 @@
...
+
+import aztech.modern_industrialization.MI;
+import aztech.modern_industrialization.MIText;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import net.minecraft.network.chat.MutableComponent;
+import net.minecraft.resources.ResourceLocation;
+
+public class HatchTypes {
```
Technici4n, line 34: These global registration functions are not great with parallel mod loading (there can be thread-safety issues, for example here the map is not thread-safe). In NeoForge we'd usually register things via an event, but maybe it's a bit too much here.

src/main/java/aztech/modern_industrialization/machines/multiblocks/HatchFlags.java
```diff
@@ -23,24 +23,32 @@
...
 public class HatchFlags {
     public static final HatchFlags NO_HATCH = new Builder().build();
 
-    private final int flags;
+    private final Set<HatchType> allowed;
 
-    public HatchFlags(int flags) {
-        this.flags = flags;
+    public HatchFlags(Set<HatchType> allowed) {
+        this.allowed = Collections.unmodifiableSet(allowed);
```
Technici4n: It's better to use `Set.copyOf` which will 1) prevent mutations inside of the passed set affecting the unmodifiable view and 2) will be a no-op if the set is already immutable.

src/main/java/aztech/modern_industrialization/machines/multiblocks/HatchType.java
```diff
@@ -23,42 +23,40 @@
...
-    FLUID_INPUT(2, MIText.FluidInputHatch),
-    FLUID_OUTPUT(3, MIText.FluidOutputHatch),
-    ENERGY_INPUT(4, MIText.EnergyInputHatch),
-    ENERGY_OUTPUT(5, MIText.EnergyOutputHatch),
-    NUCLEAR_ITEM(6, "nuclear_item_hatch"),
-    NUCLEAR_FLUID(7, "nuclear_fluid_hatch"),
-    LARGE_TANK(8, "large_tank_hatch");
+public class HatchType {
+    private final ResourceLocation id;
+    private final Either<MutableComponent, ResourceLocation> descriptionOrBlockId;
```
Technici4n: I would probably use a normal `Component` and copy it in `description`, to make sure that the mutable component doesn't get modified here.

src/main/java/aztech/modern_industrialization/machines/multiblocks/HatchType.java
```diff
@@ -23,42 +23,40 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return id.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        return o instanceof HatchType other && id.equals(other.id);
```
Technici4n, line 61: With such an equals implementation the class should be made `final` to ensure reflexivity of `equals`.

src/main/java/aztech/modern_industrialization/machines/multiblocks/HatchTypes.java
```diff
@@ -0,0 +1,85 @@
...
+    public static final HatchType ITEM_OUTPUT = create(MI.id("item_output"), MIText.ItemOutputHatch);
+    public static final HatchType FLUID_INPUT = create(MI.id("fluid_input"), MIText.FluidInputHatch);
+    public static final HatchType FLUID_OUTPUT = create(MI.id("fluid_output"), MIText.FluidOutputHatch);
+    public static final HatchType ENERGY_INPUT = create(MI.id("energy_input"), MIText.EnergyInputHatch);
+    public static final HatchType ENERGY_OUTPUT = create(MI.id("energy_output"), MIText.EnergyOutputHatch);
+    public static final HatchType NUCLEAR_ITEM = create(MI.id("nuclear_item"), MI.id("nuclear_item_hatch"));
+    public static final HatchType NUCLEAR_FLUID = create(MI.id("nuclear_fluid"), MI.id("nuclear_fluid_hatch"));
+    public static final HatchType LARGE_TANK = create(MI.id("large_tank"), MI.id("large_tank_hatch"));
+
+    public static HatchType create(ResourceLocation id, MutableComponent description) {
```
Technici4n: Since this registers it as well I would call this `register`.

src/main/java/aztech/modern_industrialization/machines/multiblocks/HatchTypes.java
```diff
@@ -0,0 +1,85 @@
...
+        var type = new HatchType(id, description);
+        registeredHatches.put(id, type);
+        return type;
+    }
+
+    public static HatchType create(ResourceLocation id, MIText description) {
+        return create(id, description.text());
+    }
+
+    public static HatchType create(ResourceLocation id, ResourceLocation blockId) {
```
Technici4n: Add a javadoc that the blockId corresponds to the user-facing description of the hatch type.

src/main/java/aztech/modern_industrialization/machines/multiblocks/HatchFlags.java
Technici4n: (Note that it is relatively newer -- added in Java 10.)

src/main/java/aztech/modern_industrialization/machines/multiblocks/HatchTypes.java
Swedz: I feel like it's a bit excessive to use a registry considering that `MachineCasing` and `CableTier` don't. If you want to move towards stuff in MI being more registry driven, I am completely on board though.

Technici4n: I think it's fine to leave this as is. Just be careful where you call it from. (E.g. from an `enqueueWork` callback if this is done from a parallel event).

Issue #1101, opened 2025-05-04
Fusion Reactor defaults to 2x helium 3 with tritium


Aphyxia:
I have 2 fluid input hatches in the fusion reactor. One with tritium and the other with helium 3. I would like it to run the tritium + helium 3 recipe but it decides to take 2x helium 3 from the one fluid input instead. How to force it to do the other recipe?

PR #1167, opened 2025-08-29
Require that output slots be locked when multiple recipes match the input

Swedz:
~~Requires that #1101 be addressed first - this causes the Fusion Reactor to be unable to run with the presence of both H3 + Tritium.~~

This addresses the issue of recipes where the inputs may be the same (or merely overlap) but the outputs are different. The way it works now is inconsistent and results in confusing behavior for some users. This change aims to make the behavior consistent and communicate to the user what is happening.

If you feel as though the visual doesn't fit or needs adjusting, let me know and I'll make the changes.

Somewhat related, but I also made a slight adjustment to the multiblock gui background texture and text alignment because it was slightly misaligned and we need at least one extra line to display the steam overclock text.

### Here's a few examples of what this accomplishes:

This assembler can make either Clean Stainless Steel Machine Casings or Stainless Steel Drill Heads. The machine won't run unless the slots are locked and it is forced to pick one. You may lock to both items to have the machine simply pick indiscriminately, as per original behavior. It's assumed if someone locks slots to overlapping recipes they are being deliberate.
image

Another example are the Distillery recipes. When the fluid is added and no slot is locked, right now, it just picks one and starts processing it.
image
image

This also supports the multiblock gui view. Here's an example of what that looks like.
image


aleck099:
@Swedz Quarries stop to work and ask for locking the output. I had no enough output slots for all the ores. This doesn't make sense since you won't mistakenly craft wrong ores with drills.

Swedz:
Does your modpack have multiple recipes for the quarry that use the same input? I will look into this today.

Swedz:
After some testing I can confirm this does work as intended, as far as I can tell. I tested each quarry recipe and all work with as few as one output slot without needing to lock slots. Please check that your pack does not add any duplicate recipes.

Swedz:
@aleck099 I believe that problem pertains to situations where multiple recipes exist and match for inputs and outputs, but the recipe conditions differ. #1284 should resolve this, assuming your recipe conditions in your pack don't also overlap.

aleck099:
> Does your modpack have multiple recipes for the quarry that use the same input? I will look into this today.

The only extra ore is zinc ore from Create.

Swedz:
@aleck099 Do you perhaps have multiple different drills in your input slots? The problem is the machine can run multiple different recipes and it wants you to choose which one. Try only putting one drill type in at a time.

aleck099:
Bronze drill only

Swedz:
@aleck099 Have you changed the recipes in the quarry? It may be you have added a new bronze drill recipe but not removed the old one, resulting in a conflict.

Swedz:
Also, I would like to point out that MI automatically includes zinc ore in the bronze drill recipe for the quarry. No need to be adding it yourself.

aleck099:
Yes, I didn't add any custom changes about mi. The zinc ore was there the first time I started the game

Swedz:
Can you run `/mi recipes check_for_conflicts` and send the result file? If it fails, make sure you have a `modern_industrialization` folder in your modpack directory and try again.

aleck099:
The problem is that when I put 2 stacks of bronze drills it says locking slots is required, but if I put only 1 stack it works

Swedz:
That was the missing piece of information I needed. This will be fixed in 2.5.2.

cbrabes:
To Clarify is this fix in 2.5.1, or will it be in 2.5.2?

Swedz:
The specific problem aleck had will be fixed in 2.5.2

PR #1170, opened 2025-09-04
Add support for data generating ores out of arbitrary strata

fmbellomy:
This PR adds data generation support for ores in dimensions other than the overworld, including adding support for ore generation based on biome tag, and making ores out of arbitrary strata.
Example:
```js
// creates Nether Tetrahedrite Ore
MIMaterialEvents.addMaterials((event) => {
  event.createMaterial("Tetrahedrite", "tetrahedrite", 0xd22300, (builder) => {
    builder
      .hardness("average")
      .materialSet("metallic")
      .addParts("dust")
      .ore(
        {
          generate: true,
          ore_set: "copper",
          biome_tag: "minecraft:is_nether",
          vein_size: 8,
          veins_per_chunk: 16,
          max_y: 64
        },
        "minecraft:netherrack"
      ).ore(
        {
          generate: true,
          ore_set: "copper",
          biome_tag: "minecraft:is_end",
          vein_size: 12,
          veins_per_chunk: 16,
          max_y: 80
        },
        "minecraft:end_stone"
      )
      .rawMetal("copper");
  });
});
```


src/main/java/aztech/modern_industrialization/compat/kubejs/material/MaterialBuilderJSWrapper.java
```diff
@@ -143,14 +146,16 @@ public MaterialBuilderJSWrapper specialCasing(String englishName, String path, f
         return this;
     }
 
-    public MaterialBuilderJSWrapper ore(JsonObject json, boolean deepslate) {
-        materialBuilder.addParts(creator.orePart(json, deepslate));
+    public MaterialBuilderJSWrapper ore(JsonObject json, Block stoneType) {
```
Swedz: `Block` is a registry entry and thus should not be referenced before or during the registry process. These references to `Block` should likely use `ResourceLocation` instead.

fmbellomy:
Updated PR to change references to Block to ResourceLocation and added support for generating ores depending on biome tag.
This implementation does still have the limitation that whatever strata the ore is meant to replace must be from the vanilla game. It might be possible to use `extra_datagen_resources` to get around this, but I wasn't able to get it to work in my testing.

Technici4n:
I did not forget, just did not have the chance to look at this PR yet.

Technici4n:
Hi @fmbellomy, I pushed some stuff to this PR, I hope that's ok. I'm happy to clarify anything. If this looks good to you we can merge it.

fmbellomy:
All your changes make sense to me. I feel it would be _convenient_ if data generation logged an error instead of crashing the game whenever `min_xp` and `max_xp` didn't match the amount expected for a given ore's main type, but I can see the case for making something that's a bug in 99% of cases as obvious as possible to users. 

I happen to be in the unfortunate 1% of cases where making a Redstone Ore block that doesn't drop experience is actually _intentional_, but I'm also pretty sure this is something that can be fixed with a KubeJS server script on block loot tables anyway.

Other than that very minor inconvenience, everything still works as expected with those changes, so I'd be satisfied with this getting merged as-is.

Technici4n:
I wouldn't want "errors" in the logs that end up being ignored. I'm not sure you can override the drops since they are hardcoded in the `OreBlock`.

Technici4n:
Note to self: re-run datagen to make sure it didn't change.

PR #1171, opened 2025-09-04
Allow namespaces on AbstractCraftingMultiblockBlockEntity

Luligabi1:
`AbstractCraftingMultiblockBlockEntity` slipped through the cracks back in #838


Technici4n:
Is there a use case for the `MachineGuiParameters.Builder` overloads? If not I would only do the RL and String ones.

Swedz:
Only reason is to make less changes to MI overall. There should be no problem with removing them.

Technici4n:
But they're new, no? So there shouldn't be anything to change in MI itself.

Swedz:
My bad, I misread the message haha. The only uses for it would be to be able to exclude the lock button or define a custom GUI height.

Luligabi1:
> Is there a use case for the `MachineGuiParameters.Builder` overloads? If not I would only do the RL and String ones.

I can see it being useful if someone plans to add more gui components above or underneath `CraftingMultiblockGui`... tbh I only did it with a "who knows, might be useful for someone else" sentiment in mind

Technici4n:
I'd rather not add unused methods as it makes further evolution more difficult especially when it's not clear that the methods have a use.

Technici4n (approved):

PR #1229, opened 2025-11-28
Update ChunkEventListeners.java

Fr3shlama:
> Remove unnecessary thread assertion that offered no real protection and replaced it with proper thread handoff.
> 
> This avoids crashing in valid multi-threaded environments (e.g. DimThreads) and lets the code run where it's actually supposed to.

This PR removes the hard thread assertion that previously caused a runtime crash whenever execution didn’t happen on the exact expected server thread—regardless of whether it was still a valid and safe execution context (such as in DimThreads or other async-capable environments).

Instead of forcefully throwing exceptions when not on the main thread, the code now properly schedules the work back onto the correct executor, which is the intended way to handle thread-sensitive operations in Minecraft.

What this fixes:
- Prevents unnecessary crashes caused by isSameThread() checks
- Improves compatibility with async/threading mods without impacting vanilla behavior

What this doesn’t do:
- It doesn’t assume there is only one “real” server thread
- It doesn’t crash on valid execution contexts that weren’t explicitly accounted for

This keeps the original intent (avoiding unsafe access) while removing the artificial limitation that made the code less compatible than it needed to be.

### DIFF
--- a/src/main/java/aztech/modern_industrialization/machines/multiblocks/world/ChunkEventListeners.java
+++ b/src/main/java/aztech/modern_industrialization/machines/multiblocks/world/ChunkEventListeners.java
@@ -78,7 +78,7 @@ public class ChunkEventListeners {
     public static void onBlockStateChange(Level world, ChunkPos chunkPos, BlockPos pos) {
         // We skip block state changes that happen outside of the server thread.
         // Hopefully that won't cause problems.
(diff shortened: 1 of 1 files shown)


Technici4n:
Dimensional threading is not supported by other parts of the mod (or most mods, for the matter), and I have no intention of changing that. In other words, there is no "valid and safe execution context" like ChatGPT claims. The `listeners` map can for example be corrupted by concurrent access from multiple threads.

PR #1235, opened 2025-12-09
Update Ukrainian guidebook

StarmanMine142:


Technici4n:
Thanks. Do the images need to be duplicated for Ukrainian? I would have thought that the common ones (for English) would be picked up.

PR #1247, opened 2026-02-26
Split insertion and extraction value tracking in pipe stats

rlnt:
This pull request fixes a niche bug in how the pipe statistics tracking works.

If an extraction and an insertion don't happen in the same tick, the effective flow rate is doubled because both operations are tracked as individual steps. My fix separates the tracking of the extraction and the insertion, and takes the higher value for the statistics value.

The bug can be reproduced by a machine that only uses energy every 2 ticks.
A more in-depth discussion about this issue can be found [on the Discord](https://discord.com/channels/754461970870173767/760125612126896196/1476543821117390974).

### DIFF
--- a/src/main/java/aztech/modern_industrialization/pipes/PipeStatsCollector.java
+++ b/src/main/java/aztech/modern_industrialization/pipes/PipeStatsCollector.java
@@ -28,16 +28,23 @@ public class PipeStatsCollector {
     private static final int REFRESH_RATE = 20;
 
     private long lastStat = 0;
-    private long currentTot = 0;
+    private long currentTotInsert = 0;
+    private long currentTotExtract = 0;
     private int ticks = 0;
 
     public void addValue(long newMoved) {
-        currentTot += newMoved;
+        addValues(newMoved, 0);
+    }
+
+    public void addValues(long extracted, long inserted) {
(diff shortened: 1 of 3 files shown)


src/main/java/aztech/modern_industrialization/pipes/PipeStatsCollector.java
```diff
@@ -28,16 +28,23 @@ public class PipeStatsCollector {
...
 
     private long lastStat = 0;
-    private long currentTot = 0;
+    private long currentTotInsert = 0;
+    private long currentTotExtract = 0;
     private int ticks = 0;
 
     public void addValue(long newMoved) {
-        currentTot += newMoved;
+        addValues(newMoved, 0);
```
Technici4n, line 36: Should this maybe be `newMoved, newMoved`?

rlnt: It could, but it doesn't matter since the logic only takes the maximum value. If you prefer that, I can do that.

Technici4n: Ah right. Good enough then.

PR #1250, opened 2026-03-11
Only include caller class of transactions when config option is true (false by default)

Swedz:
Calling `StackWalker#getCallerClass` frequently comes with a cost of performance. This adds a config option that is off by default that must be on for the caller class to be fetched.


Technici4n (approved):
Nice, thanks

PR #1314, opened 2026-07-25
Improve machine insert item performance

Swedz:
A bit messy, but it does help situations with large pipe networks transferring lots of items. This lets the insert fail without getting an `ItemVariant`, which can take some time due to a map lookup.


src/main/java/aztech/modern_industrialization/inventory/MIItemStorage.java
```diff
@@ -131,7 +134,7 @@ public boolean currentlyWhitelisted() {
         @Override
         public void getWhitelistedItems(Set<Item> whitelist) {
             for (ConfigurableItemStack stack : stacks) {
-                if (stack.pipesInsert && stack.getLockedInstance() != Items.AIR) {
+                if (stack.pipesInsert && !stack.isLockedTo(Items.AIR) && stack.getAmount() < stack.getCapacity()) {
```
Technici4n: Not sure that checking the amount is a good idea as that could change during I/O operations.

Swedz: That is a good point.

PR #1321, opened 2026-07-28
Rebalance Diesel Generators, Boilers, and the Large Steam Turbine

Swedz:
Warrants some more discussion before merging, thus this is a draft.

Changes are as follows:
- Removed Tritium and Deuterium as fluid fuels (because why?).
- Buffed Hydrogen as a fluid fuel to 2 EU per mb. Still slightly energy negative when making it directly, though (which is good IMO).
- Boilers only get 2x fuel efficiency from item fuels.
- Diesel generators get 2x fuel efficiency from fluid fuels (it can only do fluid fuels anyway, obviously).
- Large Steam Turbine is now EV tier, and costs 2 Processing Units total.
  - Before it only took Digital Circuits (HV). There was no real reason not to make it over normal HV turbines despite it taking a few more circuits. This makes it a proper upgrade from HV to EV, as well as more consistent with the Large Diesel Generator, which acts as its Diesel counterpart.
- Large Diesel Generator is still EV tier, but costs 2 Processing Units total rather than the previous 10.
image

***

Some things to consider for the future:
- Perhaps the Large Diesel Generator could get efficiency scaling depending on how much EU is actually being produced. For example, when producing <10% of what it could theoretically produce, it would get extra fuel efficiency (+20%?). Conversely, when producing >90% of what it could theoretically produce, it would get decreased fuel efficiency (-20%?).
- Perhaps the Large Diesel Generator could be made out of Stainless Steel instead of Titanium? Although, Titanium does have very few uses and this would mean it'd have even one less use.
- Maybe we make the fuel waste and heat up time from boilers less forgiving. Simply extending the time it takes a large boiler to go from 0% heat to 100% heat to a minute or longer would go a long way. As well as increasing the speed that heat is lost. This means that even with a redstone controlled system, your power output can fluctuate a lot unless you're wasting a lot of fuel. Since redstone locking doesn't preserve heat anymore. This would give even more credibility to Diesel over Boilers, since coal is really easy and has little to no downside.


James103:
Doubling all fluid fuel values will also approximately double the mining speed and attack damage of all diesel tools unless a `/ 8.0f` is replaced with `/ 16.0f` and a `/ 600` is replaced with `/ 1200` here:
https://github.com/AztechMC/Modern-Industrialization/blob/71de75d9c3516fa5a08db1fc7567f51088624e59/src/main/java/aztech/modern_industrialization/items/diesel_tools/DieselToolItem.java#L125
https://github.com/AztechMC/Modern-Industrialization/blob/71de75d9c3516fa5a08db1fc7567f51088624e59/src/main/java/aztech/modern_industrialization/items/diesel_tools/DieselToolItem.java#L134
Diesel Jetpack maximum speed will also be doubled unless a couple `/ 200.0` are replaced with `/ 400.0` here:
https://github.com/AztechMC/Modern-Industrialization/blob/71de75d9c3516fa5a08db1fc7567f51088624e59/src/main/java/aztech/modern_industrialization/items/armor/JetpackItem.java#L96-L101
Feel free to pick different divisors if you mean to actually buff these diesel-powered tools and jetpack.

James103:
The failing test is caused by Biodiesel fuel value being adjusted from 250 to 500 EU/mB.
https://github.com/AztechMC/Modern-Industrialization/blob/71de75d9c3516fa5a08db1fc7567f51088624e59/src/main/java/aztech/modern_industrialization/test/GeneratorTests.java#L56-L62

It should be adjusted to:
> Should produce 64 per tick: 448 after 7 ticks and only 500 (the biodiesel value) after 8 ticks

Swedz:
> Should produce 64 per tick: 448 after 7 ticks and only 500 (the biodiesel value) after 8 ticks

It should actually be 960 after 15 ticks and 1000 after 16 ticks, since Diesel Generators get 2x fuel efficiency with this PR.

James103:
If I understand accurately, the net effect of this change divides all fuel consumption of diesel generators by 4 while leaving their EU production unchanged, or equivalently makes each bucket of a fluid fuel produce 4x more EU than before.

For example, 100 mB/t of boosted diesel is enough to generate 80,000 EU/t before this change, 320,000 EU/t afterward.
However, the required number of Diesel Generators to extract all EU increases from 5 before this change to 20 after.

Swedz:
Effectively, yes. It's still being discussed, this is not a final draft.

MuteTiefling:
> Perhaps the Large Diesel Generator could be made out of Stainless Steel instead of Titanium? Although, Titanium does have very few uses and this would mean it'd have even one less use.

Currently, the diesel system doesn't even feel worth interacting with since coal for boilers is so easy. Moving to diesel is a massive investment and risky since it means even more complexity in the fuel chain and more potential points of failure. 

Moving it to stainless could help since you'd significantly reduce the barrier to entry. 

Maybe the large turbines could then use titanium in their place?

Swedz:
Yes, coal is really easy. The hope is to make Diesel more appealing with these changes.

With regards to titanium, to be fair, it is not that far off to obtain once you have processing units available anyway. Which these multiblocks require now.

The arguments in favor of Diesel here would be:
- It is perfectly fuel efficient out of the box. Diesel generators stop consuming fuel when they become full. There is no heat up time where you are producing less EU/t.
  - Potentially, an additional change could be to make the fuel waste and heat up time from boilers less forgiving. Since redstone locking doesn't preserve heat anymore.
- Diesel generators, while they are a similar resource cost to turbines, have double the EU/t output per turbine (excluding the multiblock). This means half as many generators and no boilers are needed for the same EU/t output.

Which for some might not be strong enough to scale up Diesel production for it to be sustainable. Considering you already have to produce the fuels required to make Diesel, I don't think it's asking too much of the player.

GW-Dev:
I personally think making boilers have a less efficient ramp up and down, coupled with the proposed diesel generator efficiency curve mechanic, would make for a far more interesting power generation system since it would add unique drawbacks to each generation type.

MuteTiefling:
> Yes, coal is really easy.

Not only is it easy, it's just more reliable. Coal only requires very simplistic bronze drills to obtain, these are easy to scale up and the lower relative complexity means if something does go wrong, it's easier to get a system back online from a blackout.

Oil, on the other hand, not only needs higher complexity drills (aluminum being a challenge to keep up with at most levels of play), but also a higher complexity processing line. So, if your system isn't properly balanced and you end up in a blackout scenario, getting back online is rough. Unless, of course, you maintain a boiler system as emergency power for the diesel system... at which point you're not really off coal.

That, for me, is the challenge to overcome. From a player's perspective, there's no reason to go to Diesel because it's such high risk. I had constant materials shortages around Aluminum, any one of which would have stopped my entire system. At least with coal, I knew I wasn't going to lose power if my aluminum production got stuck for some reason. Diesel needs to look attractive enough to try, and it needs to be reliable.

truepurpledino:
This probably isn't included here, but the MI Guide entries for the boilers and diesel gens will need to be changed to reflect these changes.

Swedz:
It's not included, good catch. I'll update that at some point before this gets merged.

src/generated/resources/assets/modern_industrialization/lang/en_us.json
```diff
@@ -1655,7 +1655,8 @@
   "text.modern_industrialization.DirectEnergy": "Direct Energy for one capture",
   "text.modern_industrialization.DirectHeatByDesintegration": "Direct Heat for one capture",
   "text.modern_industrialization.Disabled": "Disabled",
-  "text.modern_industrialization.DoubleFluidFuelEfficiency": "Double efficiency! (EU x2 per fuel used)",
+  "text.modern_industrialization.DoubleFluidFuelEfficiency": "Double efficiency! (EU x2 per fluid fuel used)",
```
Technici4n: We should maybe make this more clear. E.g. over multiple lines:
```
- Accepts item fuel at double efficiency! (EU x2 per fuel used)
- Accepts fluid fuel at normal efficiency!
```
or similar

Technici4n (approved):
LGTM. We could make the tooltips clearer I think. Also, remember to update the guidebook! ;)

src/generated/resources/assets/modern_industrialization/lang/en_us.json
Swedz: I agree, though I'm not sure how to go about it. Adding the lines via `FluidItemConsumerComponent#getTooltips` results in turbines getting the "Accepts fluid fuel at normal efficiency". Which to me seems like it would be confusing to read. 

Edit: My solution was to add a `hideEfficiencyTooltip` field to the component which is only true for turbines. Also added a method to the builder for KubeJS support. Seems simple enough to me.

Swedz:
I have updated the guide. If anyone wants to take a look and see if I missed anything, it would be much appreciated!

src/generated/resources/assets/modern_industrialization/lang/en_us.json
Swedz: I think it makes most sense to replace the "double efficiency" part with a placeholder. This is because the KubeJS plugin supports an arbitrary EU multiplier. That capability also isn't mentioned in the docs, but I will remedy that here in a moment. Below is the wording I have at the moment.

```
- Accepts item fuel at 2x efficiency! (EU 2x per item used)
- Accepts fluid fuel at 2x efficiency! (EU 2x per mb used)
- Accepts item fuel at normal efficiency.
- Accepts fluid fuel at normal efficiency.
```

image
image

docs/ADDING_MACHINES.md
```diff
@@ -382,6 +382,10 @@ builder.fluid("minecraft:lava", 10) // This will make the generator accept lava
 // multiple fuels can be added ex:
 builder.item("minecraft:coal", 100).fluid("minecraft:lava", 10);
 // will automatically add the correct input slot in the machine inventory
+
+builder.euMultiplier(2) // This will make the fuels burned in the generator be twice as efficient
```
James103, line 386: Would it work if `builder.euMultiplier(1.5)`, resulting in it burning fuel at 1.5x efficiency (150 EU generated per 100 EU of fuel)?

Swedz: No, it expects a `long` value. I could support `double`s if that is at all desirable?

James103: Support for `double` would be fine, and would enable one to have, for example, higher-tier generators have slightly reduced or enhanced efficiency.

PR #1362, opened 2026-09-14
add Fluid Pipe support for the Config Card

fmbellomy:
This is about as naive an implementation as I could get away with writing.

I added a fluid variant to the pipe config record, and the copy implementations for item and fluid pipes do their best to provide "empty" fields when pasting to their respective counterparts. 

I wasn't sure where to put the renamed `SavedPipeConfig` since I no longer felt it belonged solely in the `item` package. `api` was just my best guess for where it should live, but I'm more than willing to move it again. Lastly, most of the fluid pipe config card code was just copy/pasted (heh) directly from the item pipe config card implementation. This introduces a fair bit of code duplication, which may or may not be fine.


src/main/java/aztech/modern_industrialization/pipes/fluid/FluidNetworkNode.java
```diff
@@ -211,6 +223,31 @@ public PipeMenuProvider getConnectionGui(Direction guiDirection, PipeScreenHandl
         return null;
     }
 
+    @Override
+    public boolean customUse(PipeBlockEntity pipe, Player player, InteractionHand hand, @Nullable Direction hitDirection) {
+        for (FluidNetworkNode.FluidConnection conn : connections) {
```
Swedz: Perhaps we can add an interface such as `NetworkNodeConnection` that includes methods necessary for both this and the `customUse` implementation in `ItemNetworkNode`. The code here is practically identical to the code there.

src/main/java/aztech/modern_industrialization/pipes/fluid/FluidNetworkNode.java
```diff
@@ -231,6 +268,38 @@ private boolean canExtract() {
...
+            }
+            boolean remesh = config.connectionType() != type;
+            type = config.connectionType();
+            priority = config.insertPriority();
+            long networkAmount = 0;
+            for (var entry : network.iterateTickingNodes()) {
+                FluidNetworkNode fluidNode = (FluidNetworkNode) entry.getNode();
+                networkAmount += fluidNode.amount;
+            }
+            if (networkAmount == 0) {
```
Swedz: Instead of doing the empty check like this, perhaps we can widen access to `FluidNetwork#isEmpty(boolean)` and use that instead? You'd want to pass `false` for the `onlyFluid` parameter. This will also make sure every node for the network is loaded before changing the fluid of the network.

src/main/java/aztech/modern_industrialization/pipes/api/SavedPipeConfig.java
```diff
@@ -22,27 +22,31 @@
...
+import aztech.modern_industrialization.thirdparty.fabrictransfer.api.fluid.FluidVariant;
 import com.mojang.serialization.Codec;
 import com.mojang.serialization.codecs.RecordCodecBuilder;
 import java.util.List;
 import net.minecraft.world.item.ItemStack;
 
-public record SavedItemPipeConfig(
+public record SavedPipeConfig(
         PipeEndpointType connectionType,
+        FluidVariant fluid,
```
Swedz: Kinda nit-picky, but I would place this parameter after the `upgrade` parameter. Reason being the first 4 parameters are relevant to both fluid and item pipes. Having this here breaks that flow.

src/main/java/aztech/modern_industrialization/pipes/api/SavedPipeConfig.java
```diff
@@ -22,27 +22,31 @@
...
         int extractPriority,
         List<ItemStack> filter,
         ItemStack upgrade) {
-    public static final Codec<SavedItemPipeConfig> CODEC = RecordCodecBuilder.create(instance -> instance.group(
-            ItemNetworkNode.CONNECTION_TYPE_CODEC.fieldOf("connectionType").forGetter(SavedItemPipeConfig::connectionType),
-            Codec.BOOL.fieldOf("whitelist").forGetter(SavedItemPipeConfig::whitelist),
-            Codec.INT.fieldOf("insertPriority").forGetter(SavedItemPipeConfig::insertPriority),
-            Codec.INT.fieldOf("extractPriority").forGetter(SavedItemPipeConfig::extractPriority),
+    public static final Codec<SavedPipeConfig> CODEC = RecordCodecBuilder.create(instance -> instance.group(
+            ItemNetworkNode.CONNECTION_TYPE_CODEC.fieldOf("connectionType").forGetter(SavedPipeConfig::connectionType),
```
Swedz: This has made me realize the codec for `PipeEndpointType` is in the `ItemNetworkNode` class, which is a bit odd. Especially since this is no longer only used for item pipes. Maybe we should move it to `PipeEndpointType` and rename it to `CODEC`?

Swedz (changes requested):

src/main/java/aztech/modern_industrialization/pipes/fluid/FluidNetworkNode.java
```diff
@@ -211,7 +215,25 @@ public PipeMenuProvider getConnectionGui(Direction guiDirection, PipeScreenHandl
         return null;
     }
 
-    private class FluidConnection {
+    @Override
+    public boolean customUse(PipeBlockEntity pipe, Player player, InteractionHand hand, @Nullable Direction hitDirection) {
+        for (FluidNetworkNode.FluidConnection conn : connections) {
```
Swedz: This can be reduced further. You can add a `getDirection` method to `NetworkNodeConnection` (which of course returns the direction field value). Then add a static method to `NetworkNodeConnection` named something like `handleUse` (not sure that's a great name). Have it take in an `Iterable<NetworkNodeConnection>` parameter (and others that are needed) and do the iteration over the connections there. Then both `customUse` implementations can just call that.

Swedz (changes requested):

src/main/java/aztech/modern_industrialization/pipes/fluid/FluidNetworkNode.java
```diff
@@ -280,6 +267,10 @@ public void applyConfig(PipeBlockEntity pipe, @Nullable SavedPipeConfig config,
             }
         }
 
+        public Direction getDirection() {
```
Swedz, line 271: Please include the `@Override` annotation here

Swedz (changes requested):

Swedz:
Doing some playing with it now, and here are my thoughts:

1. I find it odd that the item pipe config saving stores the insert and extract priorities to the config, where the fluid pipe config saving stores the priority as just the insert priority.
2. It is strange that you can take a config saved from a fluid/item pipe and use it on the other. Perhaps we could include some flag that determines if the config is for fluid or item pipes. Then when its a saved fluid config, we can do nothing on item pipes (just pass, so it'd open the pipe menu). Similarly an item config could not be applied to a fluid pipe. We can ignore item 1 above if we do this.
3. The tooltips for the pipe config card need to be updated to either speak generally about pipes, or specifically say "item and fluid pipes". Right now it just says "item pipes". The keys in `MIText` should also be updated accordingly. https://github.com/AztechMC/Modern-Industrialization/blob/e0b405a1f24917db38cb5e82836511f3ad357afe/src/main/java/aztech/modern_industrialization/MIText.java#L81-L83
4. The "Configured (X items)" line should be adjusted for when its a fluid pipe config to say "Configured (\<name of fluid\> fluid)" or "Configured (no fluid)". https://github.com/AztechMC/Modern-Industrialization/blob/e0b405a1f24917db38cb5e82836511f3ad357afe/src/main/java/aztech/modern_industrialization/items/ConfigCardItem.java#L104-L116
5. The tooltip images probably could be not included when its a fluid configuration as well. Since the config would always be empty. https://github.com/AztechMC/Modern-Industrialization/blob/e0b405a1f24917db38cb5e82836511f3ad357afe/src/main/java/aztech/modern_industrialization/items/ConfigCardItem.java#L134-L140

src/main/java/aztech/modern_industrialization/items/ConfigCardItem.java
```diff
@@ -102,16 +103,28 @@ public InteractionResultHolder<ItemStack> use(Level level, Player player, Intera
...
         if (savedConfig != null) {
             var filterSize = savedConfig.filter().size();
             MutableComponent component;
-            if (filterSize == 0) {
-                component = MIText.ConfigCardConfiguredNoItems.text();
-            } else {
-                component = MIText.ConfigCardConfiguredItems.text(Component.literal("" + filterSize).setStyle(MITooltips.NUMBER_TEXT));
-            }
+            component = switch (savedConfig.configType()) {
+                case NONE -> Component.literal("");
```
Swedz, line 112: This could be `Component.empty()`. Though I am not sure I see the reason to add an empty line to the tooltip if the type is `NONE`? Shouldn't the line just get skipped in its entirety?

src/main/java/aztech/modern_industrialization/items/ConfigCardItem.java
```diff
@@ -102,16 +103,28 @@ public InteractionResultHolder<ItemStack> use(Level level, Player player, Intera
...
+                    } else {
+                        yield MIText.ConfigCardConfiguredItems.text(Component.literal("" + filterSize).setStyle(MITooltips.NUMBER_TEXT));
+                    }
+                }
+                case FLUID -> {
+                    if (savedConfig.fluid().isBlank()) {
+                        yield MIText.ConfigCardConfiguredNoFluid.text();
+                    }
+                    yield MIText.ConfigCardConfiguredFluid.text(
+                            savedConfig.fluid().getFluid().getFluidType().getDescription().getString());
```
Swedz, line 125: You can do `FluidVariantAttributes.getName(savedConfig.fluid())`. I'd use the `MITooltips.NUMBER_TEXT` style for the fluid name as well (same as the item count above).

src/main/java/aztech/modern_industrialization/items/ConfigCardItem.java
```diff
@@ -102,16 +103,28 @@ public InteractionResultHolder<ItemStack> use(Level level, Player player, Intera
...
         var savedConfig = stack.get(MIComponents.SAVED_CONFIG);
         if (savedConfig != null) {
             var filterSize = savedConfig.filter().size();
             MutableComponent component;
-            if (filterSize == 0) {
-                component = MIText.ConfigCardConfiguredNoItems.text();
-            } else {
-                component = MIText.ConfigCardConfiguredItems.text(Component.literal("" + filterSize).setStyle(MITooltips.NUMBER_TEXT));
-            }
+            component = switch (savedConfig.configType()) {
```
Swedz, line 111: This doesn't need to be a separate line for the variable definition

src/main/java/aztech/modern_industrialization/pipes/api/PipeConfigType.java
```diff
@@ -0,0 +1,61 @@
...
+
+import com.mojang.serialization.Codec;
+import com.mojang.serialization.DataResult;
+
+public enum PipeConfigType {
+    NONE(0),
+    ITEM(1),
+    FLUID(2);
+
+    private final int id;
```
Swedz, line 35: What's the point of an `id` field here? Is it just for the codec? IMO, the codec should use an all lowercase stringified value for the `PipeConfigType`, it makes the NBT data all that more readable.

If you implement `StringRepresentable`, use a string `id` field instead, you can use `StringRepresentable.fromEnum(PipeConfigType::values)` to define the codec. This tends to be how vanilla does it, and I think it's the simplest.

src/main/java/aztech/modern_industrialization/pipes/api/SavedPipeConfig.java
```diff
@@ -22,27 +22,32 @@
...
+import aztech.modern_industrialization.thirdparty.fabrictransfer.api.fluid.FluidVariant;
 import com.mojang.serialization.Codec;
 import com.mojang.serialization.codecs.RecordCodecBuilder;
 import java.util.List;
 import net.minecraft.world.item.ItemStack;
 
-public record SavedItemPipeConfig(
+public record SavedPipeConfig(
         PipeEndpointType connectionType,
+        PipeConfigType configType,
```
Swedz, line 36: I'm nitpicking again, maybe this should be the first param? It feels like the most significant aspect of the type to me so it should be first.

Swedz (changes requested):

src/main/java/aztech/modern_industrialization/pipes/api/PipeConfigType.java
fmbellomy: I will admit I pretty much just copied the implementation for PipeEndpointType, although that particular enum probably uses `int` for performance reasons.

I have changed the config type enum to use strings though, as it will indeed be much more readable.

src/main/java/aztech/modern_industrialization/items/ConfigCardItem.java
fmbellomy: It was largely just me being lazy and not wanting to wrap everything in an Optional.
I've decided to still not do that, however, and split everything out into helper functions instead.

src/main/java/aztech/modern_industrialization/pipes/api/PipeConfigType.java
```diff
@@ -0,0 +1,46 @@
...
+
+package aztech.modern_industrialization.pipes.api;
+
+import com.mojang.serialization.Codec;
+import net.minecraft.util.StringRepresentable;
+
+public enum PipeConfigType implements StringRepresentable {
+    NONE("NONE"),
+    ITEM("ITEM"),
+    FLUID("FLUID");
```
Swedz: The ids should be lowercase, that's the standard Mojang has been using

src/main/java/aztech/modern_industrialization/pipes/api/PipeConfigType.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private final String id;
+
+    PipeConfigType(String id) {
+        this.id = id;
+    }
+
+    public String getSerializedName() {
```
Swedz: Missing `@Overrides` annotation

src/main/java/aztech/modern_industrialization/items/ConfigCardItem.java
```diff
@@ -101,31 +105,49 @@ public InteractionResultHolder<ItemStack> use(Level level, Player player, Intera
...
+    private void addItemFilterTooltip(SavedPipeConfig savedConfig, List<Component> tooltipComponents) {
+        var filterSize = savedConfig.filter().size();
+        if (filterSize == 0) {
+            tooltipComponents.add(MIText.ConfigCardConfiguredNoItems.text()
+                    .withStyle(MITooltips.DEFAULT_STYLE));
+        } else {
+            tooltipComponents.add(MIText.ConfigCardConfiguredItems.text(
+                    Component.literal("" + filterSize)
+                            .setStyle(MITooltips.NUMBER_TEXT)
+                            .withStyle(MITooltips.DEFAULT_STYLE)));
```
Swedz: `DEFAULT_STYLE` should be applied to the outermost component, this mistakenly overwrites the `NUMBER_TEXT`

src/main/java/aztech/modern_industrialization/MIText.java
```diff
@@ -72,18 +72,21 @@ public enum MIText {
...
-    ConfigCardHelpItems1("Item pipe configuration:"),
-    ConfigCardHelpItems2("- Press %s + %s on an item pipe connection to save its settings in the card."),
-    ConfigCardHelpItems3("- Use %s on an item pipe connection to apply the settings from the card."),
+    ConfigCardHelpPipes1("Pipe configuration:"),
+    ConfigCardHelpPipes2("- Press %s + %s on an item or fluid pipe connection to save its settings in the card."),
+    ConfigCardHelpPipes3("- Use %s on an item or fluid pipe connection to apply the settings from the card."),
     ConfigCardNoCamouflageInInventory("No %s in inventory"),
     ConfigCardSet("Copied pipe settings to Config Card"),
     ConfigCardSetCamouflage("Copied %s camouflage to Config Card"),
+    ConfigCardWrongType("Saved config is not for this kind of pipe."),
```
Swedz: You need to run datagen (or commit the datagenned files). The only thing that should update in the PR is the en_us.json and the cache

src/main/java/aztech/modern_industrialization/items/ConfigCardItem.java
```diff
@@ -101,31 +105,49 @@ public InteractionResultHolder<ItemStack> use(Level level, Player player, Intera
...
+
+    private void addFluidFilterTooltip(SavedPipeConfig savedConfig, List<Component> tooltipComponents) {
+        if (savedConfig.fluid().isBlank()) {
+            tooltipComponents.add(MIText.ConfigCardConfiguredNoFluid.text()
+                    .withStyle(MITooltips.DEFAULT_STYLE));
+
+        } else {
+            tooltipComponents.add(MIText.ConfigCardConfiguredFluid.text(
+                    FluidVariantAttributes.getName(savedConfig.fluid()))
+                    .setStyle(MITooltips.NUMBER_TEXT)
```
Swedz: `NUMBER_TEXT` needs to be applied to the component that is the result of `FluidVariantAttributes.getName(savedConfig.fluid())`. This applies it to the outermost component and then has it overwritten by the `DEFAULT_STYLE`. The fluid name unintentionally ends up being gray instead of yellow.

Swedz (changes requested):

src/main/java/aztech/modern_industrialization/items/ConfigCardItem.java
```diff
@@ -101,31 +105,49 @@ public InteractionResultHolder<ItemStack> use(Level level, Player player, Intera
...
         var savedConfig = stack.get(MIComponents.SAVED_CONFIG);
         if (savedConfig != null) {
-            var filterSize = savedConfig.filter().size();
-            MutableComponent component;
-            if (filterSize == 0) {
-                component = MIText.ConfigCardConfiguredNoItems.text();
-            } else {
-                component = MIText.ConfigCardConfiguredItems.text(Component.literal("" + filterSize).setStyle(MITooltips.NUMBER_TEXT));
+            switch (savedConfig.configType()) {
+                case NONE -> {}
```
Swedz: You can omit the `NONE` case

Swedz (changes requested):

